PR #659, opened 2023-10-02
Improve Escape Sequences in String Literals

### LINKED ISSUE #658: Slice Lexer Aggressively Removes Backslashes in String Literals
Currently, Slice supports backslashes in string literals to allow things like:
```
[deprecated("Use \"Bar\" instead")]
struct Foo {}
```
So the lexer knows to treat the inner quotation marks as text.

However, after parsing, we remove these backslashes by just doing `replace('\\', "")` (remove all backslashes).
But this might remove backslashes the user put there intentionally.

For instance, this should reasonably be parsed down to `\hello`:
```
[deprecated("\\hello")]
```
Instead this will currently produce `hello`.

----

TL;DR we should change the backslash removal logic to only remove the first backslash of each pair of characters,
instead of removing all backslashes.

Note string literals are currently only allowed for attribute arguments.

InsertCreativityHere:
This PR fixes #658 and our escaping logic in general.
Now, whenever you type `\X` in a string literal, it will always be mapped to `X`.
`\\` -> `\`
`\"` -> `"`
`\n` -> `n`
`\8` -> `8`
Slice performs no additional processing, so we don't support special escape sequences like `\n` or `\uxxxx`...
But if a user wants to pass these through to their mapping, they still can with something like `\\n`.

### DIFF
--- a/src/parsers/slice/grammar.rs
+++ b/src/parsers/slice/grammar.rs
@@ -508,13 +508,17 @@ fn construct_attribute(
 }
 
 fn unescape_string_literal(s: &str) -> String {
+    // Flag that stores whether the next character we read is being escaped.
     let mut is_escaped = false;
     s.chars()
         .filter(|c| {
-            // If `c` is a backslash that isn't escaped, set that the next character is escaped,
-            // and return false to remove the backslash from the string.
-            is_escaped = *c == '\\' && !is_escaped;
-            !is_escaped
+            // If `c` is a backslash, and it isn't already escaped (ie: "\\"), then it is an escape character.
+            let is_escape_character = *c == '\\' && !is_escaped;
+            // Set `is_escaped` accordingly, so we know if the next character is being escaped.
+            is_escaped = is_escape_character;
+
+            // Return false for escape characters to filter them out of the string.
+            !is_escape_character
         })
         .collect()
 }
--- a/src/parsers/slice/grammar.rs
+++ b/src/parsers/slice/grammar.rs
@@ -507,6 +507,18 @@ fn construct_attribute(
     parser.ast.add_element(OwnedPtr::new(attribute))
 }
 
+fn unescape_string_literal(s: &str) -> String {
+    let mut is_escaped = false;
+    s.chars()
+        .filter(|c| {
+            // If `c` is a backslash that isn't escaped, set that the next character is escaped,
+            // and return false to remove the backslash from the string.
+            is_escaped = *c == '\\' && !is_escaped;
+            !is_escaped
+        })
+        .collect()
+}
+
 fn try_parse_integer(parser: &mut Parser, s: &str, span: Span) -> Integer<i128> {
     // Remove any underscores from the integer literal before trying to parse it.
     let sanitized = s.replace('_', "");
--- a/src/parsers/slice/grammar.lalrpop
+++ b/src/parsers/slice/grammar.lalrpop
@@ -275,7 +275,7 @@ Attribute: WeakPtr<Attribute> = {
 }
 
 AttributeArgument: String = {
-    <sl: string_literal> => sl.replace('\\', ""),
+    <sl: string_literal> => unescape_string_literal(sl),
     <i: identifier> => i.to_owned(),
 }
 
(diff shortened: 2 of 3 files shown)


pepone:
Ho would I pass the string/path `c:\new-folder`?

InsertCreativityHere:
All that Slice does is map `c:\\new-folder` to `c:\new-folder`
It's up the receiver of the string to decide what the meaning of `\n` is.

If you're writing an argument for `cs::type`, the string will end up in C#, which would see this string as being:
```
c:
ew-folder
```

If you're writing an argument that gets pass to a windows command prompt (somehow), it would see `C:\new-folder`,
since it attaches no special meaning to `\n`

src/parsers/slice/grammar.rs
```diff
@@ -507,6 +507,18 @@ fn construct_attribute(
     parser.ast.add_element(OwnedPtr::new(attribute))
 }
 
+fn unescape_string_literal(s: &str) -> String {
+    let mut is_escaped = false;
+    s.chars()
+        .filter(|c| {
```
externl, line 513: The logic is correct, but I think it's hard to follow. Maybe it would be easier to follow using an accumulator?

externl (approved):

src/parsers/slice/grammar.rs
```diff
@@ -507,6 +507,18 @@ fn construct_attribute(
...
+            // If `c` is a backslash that isn't escaped, set that the next character is escaped,
+            // and return false to remove the backslash from the string.
+            is_escaped = *c == '\\' && !is_escaped;
```
ReeceHumphreys, line 516:
```suggestion
            let is_backslash = *c == '\\';
            is_escaped = is_backslash && !is_escaped;
```

I found this filter required a bit of starting at to verify the logic. I think it could be a little easier to read if we extract `*c == '\\'` to its own variable.

ReeceHumphreys (approved):

src/parsers/slice/grammar.rs
```diff
@@ -507,6 +507,18 @@ fn construct_attribute(
     parser.ast.add_element(OwnedPtr::new(attribute))
 }
 
+fn unescape_string_literal(s: &str) -> String {
```
pepone, line 510: I wonder if we should instead report invalid escape sequences, I don't see the logic for converting `\n` to `n`, seems that we just want to support `"\\"` and `"\""`

pepone: That or keep `\n` as two characters ('\', 'n'), but just converting it to `n` seems surprising

externl: @InsertCreativityHere says that Rust issues an error.

pepone: That is not what we are doing, we are eating just the backslash and leaving the next character as is.

pepone: Python does not

```
>>> print("a\nb\c")
a
b\c
```

NodeJs seems to the same as this PR

```
> console.log("a\nb\c")
a
bc
```

externl: I think issuing an error is the best. Better that users don't mistakely think that writing `\n` will do what they want.

bernardnormier:
Is there an agreement on the behavior? Unless there is, opening a PR seems premature.

InsertCreativityHere:
> Is there an agreement on the behavior?

I had thought there was, but I guess we'll find out now.

### New Proposal

1) We keep the behavior and logic as-is.
2) But when we encounter an escape sequence other than `\\` or `\"`, we emit a warning saying:
```
warning: unknown escape sequence '<THE SEQUENCE>'
note: To type a literal backslash character, you must escape it using '\\'
```

externl:
For the sake of time, I'd rather just merge this PR like it is. It has "Node.js" behavior. It's unlikely anyone is every going to run into this.

InsertCreativityHere:
Okay, I agree, I'd really rather not spend too much time on an edge case in a niche.
Honestly, the main point of this was to fix the bug: #658.

I'll open an issue where we can discuss further changing how Slice handles escape sequences.

pepone (approved):

PR #662, opened 2023-10-16
Support Scoped Exceptions in 'throws' doc-comments Tags.

### LINKED ISSUE #661: Doc Comment `throws` can't link to Exceptions in other Modules
Currently the throws tag only accepts unscoped identifiers, there is no way to do:
```
interface MyInterface {
    /// @throws ::Test2::MyException: when it breaks
    op() throws ::Test2::MyException
}
```
instead the parser expects this:
```
interface MyInterface {
    /// @throws MyException: when it breaks
    op() throws ::Test2::MyException
}
```

----

Both could work in theory.

1) If we allow scoped identifiers, then the parser can resolve the link using that scoped identifier.

2) But even if we keep the current behavior (only unscoped identifier), we can still find the exception in the exception specification and link against that. We already check to ensure that anything in an `@throws` tag matches up with a specified exception.

Option 2 seems more convenient, but is more 'magical' than the option 1.

InsertCreativityHere:
This PR fixes #661 (and a slightly confusing doc comment).

### DIFF
--- a/tests/comment_tests.rs
+++ b/tests/comment_tests.rs
@@ -543,6 +543,38 @@ mod comments {
     }
 
     #[test]
+    fn throws_tag_can_specify_scoped_exception() {
+        // Arrange
+        let slice1 = "
+            mode = Slice1
+            module Tests
+
+            exception E {}
+        ";
+        let slice2 = "
+            mode = Slice1
+            module Other
+
+            interface Foo {
+                /// @throws Tests::E: it failed
+                op() throws Tests::E
+            }
+        ";
+
+        // Act
+        let ast = parse_multiple_for_ast(&[slice1, slice2]);
+
+        // Assert
+        let operation = ast.find_element::<Operation>("Other::Foo::op").unwrap();
+        let thrown_type = operation.exception_specification[0].definition();
+
+        let doc_comment = operation.comment().unwrap();
+        let TypeRefDefinition::Patched(documented_exception) = &doc_comment.throws[0].thrown_type else { panic!() };
+
+        assert_eq!(documented_exception, &thrown_type);
+    }
+
+    #[test]
     fn throws_tag_is_rejected_for_operations_that_do_not_throw() {
         // Arrange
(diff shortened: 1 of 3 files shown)


tests/comment_tests.rs
```diff
@@ -542,6 +542,38 @@ mod comments {
...
+    fn throws_tag_can_specify_scoped_exception() {
+        // Arrange
+        let slice1 = "
+            mode = Slice1
+            module Tests
+
+            exception E {}
+        ";
+        let slice2 = "
+            mode = Slice1
```
ReeceHumphreys, line 555:
```suggestion
```

Is this suppose to be a slice2 "file"?

ReeceHumphreys: If not maybe we should rename the variable

InsertCreativityHere: The `1` and `2` suffixes are for just for counting the files. Sometimes, one file is `Slice1` and another is `Slice2`, so we match them up for readability, but that's just a fluke. For example, check out: https://github.com/icerpc/slicec/blob/d5a3a9efbd548f69ac930814a8a949f4b669acc8/tests/scope_resolution_tests.rs#L85 We have `slice1`, `slice2`, and `slice3`, even though all the strings are in `Slice2` mode.

ReeceHumphreys: Hmm, not a big fan of that. I find that naming convention not great since in a test you might need some mode=slice1 and mode=slice2 in the same test for example which then I would expect to be named `slice1` and `slice2`.

Anyway, thats not related to your PR so I wouldnt worry about it for now.

ReeceHumphreys (approved):

tests/comment_tests.rs
InsertCreativityHere: I agree, these names aren't great. I think they existed before we adopted Slice-driven encoding (it `mode`).
You're welcome to open an issue with better names. Wouldn't even be hard to change, just tedious.

Issue #673, opened 2023-11-19
Refactor proxy-as-data-type


bernardnormier:
Currently, when you define a Slice interface X, you get a data type that you can encode/decode - and for example use as a struct field or operation parameter. This data type is a "proxy" to a service that implements X.

This is problematic for a number of reasons:
a) Unlike other user-defined types, the encoded representation of this type is not intuitive.  You see an enum, you have some idea of what an enum parameter looks like "over the wire". The same isn't true for a proxy.

b) The mapping for interfaces is RPC-specific, but the encoding/decoding of data types is not, so we can't have the clean-separation:
  - data types => RPC-independent
  - interfaces => RPC dependent

since defining an interface automatically defines a proxy data type.

c) For many applications, this Slice-level proxy data type is never used. It's not that common to pass a proxy as a parameter or  return value. 

d) Passing proxies around in Slice operations may not be a good design choice. And arguably this feature (any Slice interface gives you automatically a proxy data type) encourages this design.

## Proposal

a) Defining an interface no longer defines a Slice-level encodable proxy data type.

For example, the following definition would no longer compile:
```slice
interface Widget {} // no longer a data type

interface WidgetFactory {
    create() -> Widget // error
}
```

b) Allow the user to "opt-in" the old behavior with a custom type. For example:
```slice
module FooBar

interface Widget {}

// Encodes / decodes WidgetProxy as a service address; its C# API is FooBar.WidgetProxy
[cs::type("FooBar.WidgetProxy")]
custom WidgetProxy

interface WidgetFactory {
    create() -> WidgetProxy // returns a regular custom data type
}
```

It's just a regular custom data type that maps in C# to WidgetProxy.

The only special feature is as long as the user calls this custom type *InterfaceName*Proxy, he doesn't need to provide the Encode and Decode methods. slicec generates them automatically alongside the proxy struct for this interface.

This keeps the convenience of using proxy as data types, but makes this convenience opt-in. It also provides a clean separation between Slice data type (fully RPC-independent) and interfaces (RPC-dependent).

Note that it's a breaking  change can since you need to:
i) add this custom type to your Slice file (the opt-int part), and
ii) call this custom type NameProxy - not Name

InsertCreativityHere:
Sounds fine to me. This shouldn't cost any functionality, and will finally finish separating protocol constructs out of Slice.
The only downsides I see are slight:
- Passing proxies is a little more advanced now (which is fine)
- We add a _slightly_ special case to how custom types work.

----

[...]

PR #675, opened 2023-11-22
Interfaces can no longer be used as types.

InsertCreativityHere:
This implements the parser side of #673.

Interfaces can no longer be used as a type in Slice; i.e. it no longer implements the `Type` trait.

On a more conceptual level. Defining an interface no longer implicitly defines a corresponding proxy type.

### DIFF
--- a/src/patchers/encoding_patcher.rs
+++ b/src/patchers/encoding_patcher.rs
@@ -36,8 +36,11 @@ pub unsafe fn patch_ast(compilation_state: &mut CompilationState) {
                 exception_ptr.borrow_mut().supported_encodings = Some(encodings);
             }
             Node::Interface(interface_ptr) => {
-                let encodings = patcher.get_supported_encodings_for(interface_ptr.borrow());
-                interface_ptr.borrow_mut().supported_encodings = Some(encodings);
+                // We still compute which encodings an interface can be used with, so we can validate that the interface
+                // is supported by whatever compilation mode it has been declared in.
+                // But since interfaces aren't a type, we don't need to store the result of this computation, since
+                // nothing can reference this interface except other interfaces.
+                patcher.get_supported_encodings_for(interface_ptr.borrow());
             }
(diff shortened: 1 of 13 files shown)


src/patchers/encoding_patcher.rs
```diff
@@ -36,8 +36,11 @@ pub unsafe fn patch_ast(compilation_state: &mut CompilationState) {
                 exception_ptr.borrow_mut().supported_encodings = Some(encodings);
             }
             Node::Interface(interface_ptr) => {
-                let encodings = patcher.get_supported_encodings_for(interface_ptr.borrow());
-                interface_ptr.borrow_mut().supported_encodings = Some(encodings);
+                // We still compute which encodings an interface can be used with, so we can validate that the interface
+                // is supported by whatever compilation mode it has been declared in.
```
externl, line 40:
```suggestion
                // We compute which encodings an interface can be used with so we can validate that the interface
                // is supported by whatever compilation mode it has been declared in.
```

src/patchers/encoding_patcher.rs
```diff
@@ -36,8 +36,11 @@ pub unsafe fn patch_ast(compilation_state: &mut CompilationState) {
...
+                // But since interfaces aren't a type, we don't need to store the result of this computation, since
+                // nothing can reference this interface except other interfaces.
```
externl, line 42: The C# code generator uses this feature

https://github.com/icerpc/icerpc-csharp/blob/main/tools/slicec-cs/src/generators/dispatch_generator.rs#L45

externl (approved):
Looks good. See comment.

src/patchers/encoding_patcher.rs
InsertCreativityHere: You're correct. I added a function to `Interface` to still expose its supported encodings.

ReeceHumphreys (approved):
Looks good.

Issue #681, opened 2023-12-08
Add Support for `compact` Enums


InsertCreativityHere:
With the introduction of enums with associated fields, we also added the ability for enumerators to hold tagged fields.
Like we do with `struct`s, it would be good to support marking `enum`s as compact, for users who want to opt-out of the extensibility provided by tags, in exchange for saving 1 byte by omitting `TagEndMarker`.

InsertCreativityHere:
This was implemented in #686.

PR #686, opened 2023-12-12
Add Support for Compact Enums

InsertCreativityHere:
This PR implements #681 by adding support for compact enums.
Compact enums work just like compact structs: we don't encode `TagEndMarker`, and tags cannot be used within the enum.

A compact enum cannot have an underlying type (since they're already 'compact') nor can it be unchecked (since we already don't encode `TagEndMarker` then).

### DIFF
--- a/src/diagnostics/errors.rs
+++ b/src/diagnostics/errors.rs
@@ -90,6 +90,14 @@ pub enum Error {
         enum_identifier: String,
     },
 
+    /// A type was marked 'compact' when it was invalid to do so.
+    CannotBeCompact {
+        /// The kind of type that was marked compact.
+        kind: &'static str,
+        /// The identifier of the type.
+        identifier: String,
+    },
+
     /// An enumerator was found that was out of bounds of the underlying type of the parent enum.
     EnumeratorValueOutOfBounds {
         /// The identifier of the enumerator.
@@ -137,9 +145,6 @@ pub enum Error {
     /// Compact structs cannot be empty.
     CompactStructCannotBeEmpty,
 
-    /// Compact structs cannot contain tagged fields.
-    CompactStructCannotContainTaggedFields,
-
     // ----------------  Tag Errors ---------------- //
     /// A duplicate tag value was found.
     CannotHaveDuplicateTag {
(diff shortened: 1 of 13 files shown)


src/validators/enums.rs
```diff
@@ -173,3 +175,53 @@ fn cannot_contain_explicit_values(enum_def: &Enum, diagnostics: &mut Diagnostics
...
+fn check_compact_modifier(enum_def: &Enum, diagnostics: &mut Diagnostics) {
+    if enum_def.is_compact {
+        if let Some(underlying) = &enum_def.underlying {
+            Diagnostic::new(Error::CannotBeCompact {
+                kind: enum_def.kind(),
+                identifier: enum_def.identifier().to_owned(),
+            })
+            .set_span(enum_def.span())
+            .add_note(
+                "compact enums cannot also have underlying types\ntry removing either the 'compact' modifier, or the underlying type",
```
externl, line 189: I think we should try to avoid using newlines in these notes.

externl (approved):

src/diagnostics/errors.rs
```diff
@@ -365,8 +376,9 @@ implement_diagnostic_functions!(
     ),
     (
         "E018",
-        CompactStructCannotContainTaggedFields,
-        "tagged fields are not supported in compact structs\nconsider removing the tag, or making the struct non-compact"
+        CompactTypeCannotContainTaggedFields,
+        format!("tagged fields are not supported in compact {kind}s\nconsider removing the tag, or making the {kind} non-compact"),
+        kind
```
ReeceHumphreys, line 381: I agree with @externl on the new lines in the error descriptions. Do we do this anywhere else?

ReeceHumphreys (approved):

src/validators/enums.rs
InsertCreativityHere: Okay, I'll change them all to other punctuation.

What's the problem with newlines though?
The console output will already have newlines from snippets, and the json output is all string-escaped.

src/diagnostics/errors.rs
InsertCreativityHere: There's a few scattered around notes and error messages.
Just whenever something seemed more readable on two lines.

Issue #682, opened 2023-12-08
Associated fields in Enumerator


bernardnormier:
The associated fields of an Enumerator are an:
```rs
 pub associated_fields: Option<Vec<WeakPtr<Field>>>,
```

And apparently, when slicec parses an enumerator, it:
 - parses `E()` as a non-null empty vector of fields
 - parses `E` as a null vector of fields

This seems overcomplicated. Why not simply:
```rs
 pub associated_fields: Vec<WeakPtr<Field>>,
```

with an empty vec in the common situation where the enumerator has no associated field.

InsertCreativityHere:
This change was implemented in #685

bernardnormier:
This change meaning:
> Changes fields to return a Vec<Field> instead of an Option<Vec<Field>> for easier use (None => vec![]).

PR #687, opened 2023-12-13
Add Result Type Support

InsertCreativityHere:
This PR adds support for a new generic primitive type named `Result`.
It is syntactically equivalent to a `Dictionary`, but uses the `Result` keyword instead.
```
unchecked enum MyError {}

typealias StringResult = Result<String, MyError>

interface I {
    getBytes() -> Result<Sequence<uint8>, MyError>
}
```

`Result` is a Slice2 only feature. It can be used anywhere a type can be.
It is logically equivalent to a compact enum with 2 variants for 'ok' and 'error'.

Internally in the compiler, I named it `ResultType`, since `Result` is a heavily used type in Rust.

----

## Deviations from the original proposal (#682)

In the original proposal, we wanted to support tuples in the `Ok` type:
```
Result<(a: bool, tag(2), b: string?), MyError>
```
However this implementation does not support this, since we never added support for first-class tuples to Slice.
Both type parameters for `Result` must be Slice types.

We can always change this in the future, but note there is no lack of functionality.
Users must simply convert their tuple into a struct, and then that struct can be used.


externl (approved):
Looks good to me!

externl:
An action is failing with:

>`single_line_let_else_max_width` cannot have a value that exceeds `max_width`. `single_line_let_else_max_width` will be set to the same value as `max_width`

tests/optional_tests.rs
```diff
@@ -467,6 +467,72 @@ mod optional {
...
+            let field = ast.find_element::<Field>("Test::S::a").unwrap();
+            assert!(field.data_type.is_optional);
+
+            let Types::ResultType(result_type) = field.data_type().concrete_type() else { panic!() };
+            assert!(!result_type.ok_type.is_optional);
+            assert!(!result_type.err_type.is_optional);
+        }
+
+        #[test]
+        fn results_with_optional_keys_are_parsed_correctly() {
```
pepone, line 493: Should be `optional_ok`?

tests/optional_tests.rs
```diff
@@ -467,6 +467,72 @@ mod optional {
...
+            let field = ast.find_element::<Field>("Test::S::a").unwrap();
+            assert!(!field.data_type.is_optional);
+
+            let Types::ResultType(result_type) = field.data_type().concrete_type() else { panic!() };
+            assert!(result_type.ok_type.is_optional);
+            assert!(!result_type.err_type.is_optional);
+        }
+
+        #[test]
+        fn results_with_optional_values_are_parsed_correctly() {
```
pepone, line 515: `optinal_err`?

pepone (approved):

tests/optional_tests.rs
InsertCreativityHere: Yes, good catch!
I copied the dictionary tests, a little too sloppily apparently.

tests/optional_tests.rs
InsertCreativityHere: Same.

PR #688, opened 2023-12-13
Allow Settings Explicit Discriminants for Enums with Fields

InsertCreativityHere:
This PR allows users to set explicit discriminants for their enumerators, even if they have fields.
It also fixes the range of these discriminants so they must fit within [0, int32::MAX].

### DIFF
--- a/src/validators/enums.rs
+++ b/src/validators/enums.rs
@@ -14,14 +14,9 @@ pub fn validate_enum(enum_def: &Enum, diagnostics: &mut Diagnostics) {
     check_compact_modifier(enum_def, diagnostics);
     compact_enums_cannot_contain_tags(enum_def, diagnostics);
 
-    // If the enum wasn't defined in a Slice1 file, validate whether fields or explicit values are allowed,
-    // based on whether it has an underlying type. Fields in Slice1 files are already rejected by `encoding_patcher`.
-    if !enum_def.supported_encodings().supports(Encoding::Slice1) {
-        if enum_def.underlying.is_some() {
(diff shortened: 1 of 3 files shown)


src/validators/enums.rs
```diff
@@ -65,13 +59,15 @@ fn backing_type_bounds(enum_def: &Enum, diagnostics: &mut Diagnostics) {
...
+                if let Some(bounds) = underlying_type.numeric_bounds() {
+                    check_bounds(enum_def, bounds, diagnostics);
                 }
             }
             None => {
-                // No underlying type, the default is varint32 for Slice2.
-                check_bounds(enum_def, &Primitive::VarInt32, diagnostics);
+                // For enumerators in Slice2, values must fit within varint32 and be positive.
+                let varint32_max = Primitive::VarInt32.numeric_bounds().unwrap().1;
+                check_bounds(enum_def, (0, varint32_max), diagnostics);
```
bernardnormier, line 70: I find this varint32_max odd. Why define this max since it's == int32_max?

bernardnormier (approved):

src/validators/enums.rs
InsertCreativityHere: We definitely could.
I just thought it seemed cleaner to get the max from the actual Slice type, instead of the mapped type.
Using `i32::MAX` is obviously simpler though.

InsertCreativityHere: I switched to using `i32::MAX`.

externl (approved):

ReeceHumphreys (approved):
Looks good.

PR #689, opened 2023-12-18
Improved Cycle Detection Logic

### LINKED ISSUE #683: Recursive enum
The Slice parser should reject recursive enums. It currently accepts them.

For example:

```slice
enum Color { Red, White, Blue, Other(replacement: Color) }
```

InsertCreativityHere:
This PR fixes and improves the cycle detection logic.
1) Currently, the detector doesn't check inside of sequences, dictionaries, or results, so `struct S { s: Sequence<S> }` won't be caught.
This PR improves the detector to check these kinds of things.

2) Adds a path for the detector to check enums with fields for cycles.
(Currently it does not).

3) It adds logic to avoid outputting duplicate cycles:
`A -> B -> A` and `B -> A -> B` are _really_ the same cycle, no point in outputting both.
We avoid duplicates by storing sets of cycle elements like `{A, B}`.
If we detect a cycle, and it's set of elements has already been reported, we don't report it again.

Fixes #683.

### DIFF
--- a/src/validators/cycle_detection.rs
+++ b/src/validators/cycle_detection.rs
@@ -2,69 +2,179 @@
 
 use crate::ast::node::Node;
 use crate::ast::Ast;
-use crate::diagnostics::{Diagnostic, Diagnostics, Error};
-use crate::grammar::{Container, Field, Member, Types};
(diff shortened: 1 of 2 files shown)


externl (approved):
Looks good!

PR #694, opened 2024-02-21
Ordered Compilation

InsertCreativityHere:
Currently, the ordering of files is non-deterministic.
ie. When you pass `slicec` a set of files to compile, the order you pass them in is completely discarded.
This PR changes this behavior, so `slicec` preserves the order of files; this makes compilation more deterministic, and repeatable.

It does this at two levels:
- When performing file resolution (so taking command line input and finding the files on disk), we use a `HashMap`. Now we use a `Vec`. So we will resolve files in the order that the user supplied them.
- The `files` field of `CompilationState` is a `HashMap`, now it's a `Vec`. So we will parse the files in the order they were resolved, we will validate the files in the order they were resolved, etc. (The keys of the map were just `file.relative_path`, so it was also just redundant information... the path is already stored in the `SliceFile` itself.)

The only downside is we now have to use `Vec::iter().find()` when looking for a file, instead of `HashMap::get()`.

### DIFF
--- a/src/compilation_state.rs
+++ b/src/compilation_state.rs
@@ -5,13 +5,12 @@ use crate::diagnostic_emitter::{emit_totals, DiagnosticEmitter};
 use crate::diagnostics::{get_totals, Diagnostic, Diagnostics};
 use crate::slice_file::SliceFile;
 use crate::slice_options::SliceOptions;
-use std::collections::HashMap;
 
 #[derive(Debug, Default)]
 pub struct CompilationState {
     pub ast: Ast,
     pub diagnostics: Diagnostics,
-    pub files: HashMap<String, SliceFile>,
+    pub files: Vec<SliceFile>,
 }
 
 impl CompilationState {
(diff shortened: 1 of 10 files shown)


src/compilation_state.rs
```diff
@@ -5,21 +5,20 @@ use crate::diagnostic_emitter::{emit_totals, DiagnosticEmitter};
 use crate::diagnostics::{get_totals, Diagnostic, Diagnostics};
 use crate::slice_file::SliceFile;
 use crate::slice_options::SliceOptions;
-use std::collections::HashMap;
 
 #[derive(Debug, Default)]
 pub struct CompilationState {
     pub ast: Ast,
     pub diagnostics: Diagnostics,
-    pub files: HashMap<String, SliceFile>,
```
externl, line 14: Might be worth a comment to say why we're using a vec

externl (approved):

ReeceHumphreys (approved):
Looks good!

PR #696, opened 2024-05-30
Sha256 hashing for SliceFile added

ReeceHumphreys:
This PR adds a `compute_sha256_hash` method to `SliceFile`. This function is used to generate a hash that will be stable across releases. Given my comment below, should we re-export `sha2` from `slicec` since I want to use it in `slicec-cs`? This would guarantee the version `slicec-cs` uses will always be identical to the slicec version.

### DIFF
--- a/src/slice_file.rs
+++ b/src/slice_file.rs
@@ -168,17 +168,39 @@ impl SliceFile {
         formatted_snippet + &line_prefix
     }
 
-    /// Hash the combination of the filename, relative path, and raw text using a SHA-256 hash.
+    /// Hash the combination of the filename and the raw text using a SHA-256 hash.
     ///
     /// # Returns
-    /// The SHA-256 hash as a vector of bytes.
-    pub fn compute_sha256_hash(&self) -> Vec<u8> {
+    /// The SHA-256 hash as a 32-byte array.
+    pub fn compute_sha256_hash(&self) -> [u8; 32] {
         Sha256::new()
             .chain_update(self.filename.as_bytes())
-            .chain_update(self.relative_path.as_bytes())
             .chain_update(self.raw_text.as_bytes())
             .finalize()
-            .to_vec()
+            .into()
+    }
+}
+
+pub trait SliceFileCollectionExt {
+    /// Hashes all the slice files in the collection using a SHA-256 hash and returns the hexadecimal representation.
+    ///
+    /// # Returns
+    /// The SHA-256 hash as a hexadecimal string.
+    fn hash_all(&self) -> String;
+}
+
+impl SliceFileCollectionExt for &[SliceFile] {
+    fn hash_all(&self) -> String {
+        format!(
+            "{:032x}",
+            self.into_iter()
+                .map(|slice_file| slice_file.compute_sha256_hash())
+                .fold(Sha256::new(), |mut hasher, file_hash| {
+                    hasher.update(file_hash);
+                    hasher
+                })
+                .finalize()
+        )
     }
 }
 
--- a/src/slice_file.rs
+++ b/src/slice_file.rs
@@ -167,18 +167,6 @@ impl SliceFile {
 
         formatted_snippet + &line_prefix
     }
-
-    /// Hash the combination of the filename and the raw text using a SHA-256 hash.
-    ///
-    /// # Returns
-    /// The SHA-256 hash as a 32-byte array.
-    pub fn compute_sha256_hash(&self) -> [u8; 32] {
-        Sha256::new()
-            .chain_update(self.filename.as_bytes())
-            .chain_update(self.raw_text.as_bytes())
-            .finalize()
-            .into()
-    }
 }
 
 pub trait SliceFileHashable {
--- a/src/slice_file.rs
+++ b/src/slice_file.rs
@@ -4,6 +4,7 @@ use crate::grammar::*;
 use crate::utils::ptr_util::WeakPtr;
 use console::style;
 use serde::Serialize;
+use sha2::{Digest, Sha256};
 use std::cmp::{max, min, Ordering};
 use std::fmt::{Display, Write};
 
@@ -165,6 +166,19 @@ impl SliceFile {
         }
 
         formatted_snippet + &line_prefix
+    }
+
+    /// Hash the combination of the filename, relative path, and raw text using a SHA-256 hash.
+    ///
+    /// # Returns
+    /// The SHA-256 hash as a vector of bytes.
+    pub fn compute_sha256_hash(&self) -> Vec<u8> {
+        Sha256::new()
+            .chain_update(self.filename.as_bytes())
+            .chain_update(self.relative_path.as_bytes())
+            .chain_update(self.raw_text.as_bytes())
+            .finalize()
+            .to_vec()
     }
 }
 
--- a/src/slice_file.rs
+++ b/src/slice_file.rs
@@ -174,7 +174,12 @@ pub trait SliceFileHashable {
     fn compute_sha256_hash_as_bytes(&self) -> [u8; 32];
 
     /// Hashes the SliceFile using a SHA-256 hash and returns the hash as a hex string.
-    fn compute_sha256_hash(&self) -> String;
+    fn compute_sha256_hash(&self) -> String {
+        self.compute_sha256_hash_as_bytes()
+            .iter()
+            .map(|byte| format!("{:02x}", byte))
+            .collect()
+    }
 }
 
 impl SliceFileHashable for SliceFile {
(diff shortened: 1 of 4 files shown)


ReeceHumphreys:
Note: The idea with this PR is that it will enable me to do something like this in `slicec-cs`:

```rust
// Hash the vector of slice files using the sha256 algorithm.
let final_hash = compilation_state
    .files
    .iter()
    .map(|file| file.compute_sha256_hash())
    .fold(Sha256::new(), |mut hasher, file_hash| {
        hasher.update(file_hash);
        hasher
    })
    .finalize()
    .map(|byte| format!("{:02x}", byte))
    .into_iter()
    .collect::<String>();
```

src/slice_file.rs
```diff
@@ -166,6 +167,19 @@ impl SliceFile {
...
+    ///
+    /// # Returns
+    /// The SHA-256 hash as a vector of bytes.
+    pub fn compute_sha256_hash(&self) -> Vec<u8> {
+        Sha256::new()
+            .chain_update(self.filename.as_bytes())
+            .chain_update(self.relative_path.as_bytes())
+            .chain_update(self.raw_text.as_bytes())
+            .finalize()
+            .to_vec()
```
InsertCreativityHere, line 181: This is like, pretty minor, but I'd recommend using `into_array()`.
- `to_vec` will make an unnecessary copy
- The result of this hash is always a fixed size array, `to_vec` is only there for if you specifically need a `Vec` for some reason.
```suggestion
            .into_array()
```

----

The `GenericArray` type it returns is an unfortunate holdover from before Rust supported `[T; N]` syntax (it used to be that `N` needed to be an integer literal).
But it should be strictly equivalent to `[u8; 32]` (sha256 / 8 = 32 bytes)
which `into_array` should give you.

I haven't messed with this crate since I was messing around with file-system stuff, so if I'm mistaken, and this is somehow a hassle, it's fine to leave a `Vec` too.

InsertCreativityHere (commented):
Code is fine, but I'm wondering, maybe we should do this in `slicec-cs` instead?

I'm just thinking ahead, but when we port `slicec-cs` to be written in `C#`, this will become dead code. So maybe it's better to just put it in `slicec-cs` to begin with?

Also, from my understanding, only `slicec-cs` needs this logic anyways, so maybe it's best put next to where it will be used.

----

`SliceFile` is a purely public struct with purely public fields, so anything `slicec` can access, `slicec-cs` can access too.

InsertCreativityHere:
> The idea with this PR is that it will enable me to do something like this in

All good! Just wanted to say that the array returned by `finalize` can also be directly hex-ified though.
Not sure what the exact format is, but try printing the result itself (with `{:032x}`) and see if it looks reasonable.
If it does, then the last 3 functions in your chain can probably just removed.
```
    .finalize()
    .map(|byte| format!("{:02x}", byte))
    .into_iter()
    .collect::<String>();
```
But maybe it will output things in a way we don't like. I don't remember, it's been a while since I've done hashing in Rust.
Worth a try though IMO!

----

Your `02` changed to `032` in mine because a single byte takes up 2 digits,
whereas an SHA256 hash should take up 32 digits.

src/slice_file.rs
```diff
@@ -166,6 +167,19 @@ impl SliceFile {
...
     }
+
+    /// Hash the combination of the filename, relative path, and raw text using a SHA-256 hash.
+    ///
+    /// # Returns
+    /// The SHA-256 hash as a vector of bytes.
+    pub fn compute_sha256_hash(&self) -> Vec<u8> {
+        Sha256::new()
+            .chain_update(self.filename.as_bytes())
+            .chain_update(self.relative_path.as_bytes())
```
externl, line 178: Does relative path not include the file name?

externl: It's not obvious to me if we should only include the file contents and not include file name/ path.

externl:
I think it makes sense to have this in `slicec` with a helper function to hash a list of files. In the future we may want to include  this "compilation hash" as part of the serialized AST.

src/slice_file.rs
ReeceHumphreys: Relative path is just where it looks for files for example:

`relative_path` /"Greeter/" 
`file_name`: /"/Users/reece/Developer/icerpc/icerpc-csharp/examples/slice/Greeter/slice/Greeter.slice/"

We could then omit relative_path

ReeceHumphreys: I if the user has a file foo.slice and renames it to bar.slice the hash should different.

InsertCreativityHere: > Relative path is just where it looks for files

Is that an actual example? Because if so that _seems_ like a bug.
I believe that `relative_path` should fully include the `filename` here.
At the very least, maybe we need to change the field name to something more accurate...

InsertCreativityHere:
> It's not obvious to me if we should only include the file contents and not include file name/ path.

Yes, an important question is: "what are we trying to measure here?"
Without an answer to that, it's not clear what we should and shouldn't be taking here.
I'd assumed I'd missed a discussion on this, but maybe not.

InsertCreativityHere:
> In the future we may want to include this "compilation hash" as part of the serialized AST.

Sure, it can live in `slicec`, but I would be surprised if it's doesn't end up being dead code.
Even if we decide to include a hash alongside the serialized AST, it would probably include more information that the hash we're computing for our telemetry stuff.

And moving it would solve the concern of crate versions falling out of sync.
Because I would really prefer not to re-export `sha2` from `slicec`.
Adding the entire crate to our public API, for this very niche function seems pretty extreme to me.

ReeceHumphreys:
I moved what was originally going to be in `slicec-cs` into a helper function here instead. I implemented it just for slices of `SliceFile`. We could do it more generically but its not necessary imo.

externl:
> > In the future we may want to include this "compilation hash" as part of the serialized AST.
> 
> Sure, it can live in `slicec`, but I would be surprised if it's doesn't end up being dead code. Even if we decide to include a hash alongside the serialized AST, it would probably include more information that the hash we're computing for our telemetry stuff.

I'm not really following what yo mean here. Once of the things we can include in the response would be a hash of the src + reference files. No need for each language to try and compute (read all the files and figure what what's needed) this itself. We'll likely want to try and exclude unused reference files.
 
> And moving it would solve the concern of crate versions falling out of sync. Because I would really prefer not to re-export `sha2` from `slicec`. Adding the entire crate to our public API, for this very niche function seems pretty extreme to me.

- sha2 is not really niche
- we defiantly shouldn't export the sha2 crate. We should just return a string.

externl:
I think it's fine to just include
- name
- contents
No path.

src/slice_file.rs
```diff
@@ -166,6 +167,41 @@ impl SliceFile {
 
         formatted_snippet + &line_prefix
     }
+
+    /// Hash the combination of the filename and the raw text using a SHA-256 hash.
+    ///
+    /// # Returns
+    /// The SHA-256 hash as a 32-byte array.
+    pub fn compute_sha256_hash(&self) -> [u8; 32] {
```
externl, line 175: It would be nice if these two hashing functions had the same signature and return type.

tests/files/mod.rs
```diff
@@ -1,3 +1,31 @@
...
 mod io;
+
+use slicec::diagnostics::Diagnostics;
+use slicec::slice_file::SliceFileHashable;
+use slicec::slice_options::SliceOptions;
+use slicec::utils::file_util::resolve_files_from;
+use std::path::PathBuf;
+
+/// This test is used to verify that the `hash_all` method returns a hash that is stable across releases and uses
+/// the correct ordering of files.
```
externl, line 12: It would be worth testing the reverse order is also the same. No need to hardcode the hash. Just test the hash is the same in both orders.
```suggestion
///  Test that `compute_sha256_hash` returns a hash that is order independent.
```

externl (approved):

src/slice_file.rs
```diff
@@ -168,6 +169,67 @@ impl SliceFile {
     }
 }
 
+pub trait SliceFileHashable {
+    /// Hashes the SliceFile using a SHA-256 hash and returns the 32-byte array.
+    fn compute_sha256_hash_as_bytes(&self) -> [u8; 32];
+
+    /// Hashes the SliceFile using a SHA-256 hash and returns the hash as a hex string.
+    fn compute_sha256_hash(&self) -> String;
```
InsertCreativityHere: Both implementations of this are identical, so it's better to just define the function once:
```suggestion
    fn compute_sha256_hash(&self) -> String {
        self.compute_sha256_hash_as_bytes()
            .iter()
            .map(|byte| format!("{:02x}", byte))
            .collect()
    }
```

src/slice_file.rs
```diff
@@ -168,6 +169,67 @@ impl SliceFile {
...
+    /// Hash the combination of the filename and the raw text using a SHA-256 hash.
+    ///
+    /// # Returns
+    /// The SHA-256 hash as a hex string.
+    fn compute_sha256_hash(&self) -> String {
+        self.compute_sha256_hash_as_bytes()
+            .iter()
+            .map(|byte| format!("{:02x}", byte))
+            .collect()
+    }
```
InsertCreativityHere: And then you can delete these duplicates:
```suggestion
    }
```

src/slice_file.rs
```diff
@@ -168,6 +169,67 @@ impl SliceFile {
...
```
InsertCreativityHere: Same:
```suggestion
    }
```

src/slice_file.rs
```diff
@@ -168,6 +169,67 @@ impl SliceFile {
     }
 }
 
+pub trait SliceFileHashable {
```
InsertCreativityHere, line 172: I think using the extension trait pattern is a little overkill here.

It makes sense when you have a handful of types which are going to implement this, but we'll only ever have 2: `SliceFile` and `[SliceFile]`. And nothing else outside this crate will ever implement it either.

I think it'd be cleaner to just make these normal functions on `SliceFile`:
```
compute_sha256_hash(&self) -> [u8; 32] { ... }
compute_sha256_hash_for_all(files: &[Self]) -> [u8; 32] { ... }
```

And then put a `convert_to_hex(&[u8]) -> String` in util or something.
Converting a byte array to a hex string isn't specific to slice files.
That's just a general purpose utility function.

----

Also, do we need to convert the bytes to hex though?
It's all going to be bytes across the wire, and sending the byte array directly is half the size, plus then we don't need the logic to convert it.

I'm fine keeping the hex! Hex will definitely make your test easier to write.
It's just a thought.

src/slice_file.rs
```diff
@@ -168,6 +169,52 @@ impl SliceFile {
...
+            .finalize()
+            .into()
+    }
+}
+
+impl SliceFileHashable for &[SliceFile] {
+    fn compute_sha256_hash_as_bytes(&self) -> [u8; 32] {
+        // Sort the slice files by their filename before hashing them.
+        let mut sorted = self.iter().collect::<Vec<_>>();
+        sorted.sort_by(|a, b| a.filename.cmp(&b.filename));
```
InsertCreativityHere, line 203: Why do we sort the files here?
Is there a benefit to doing this?

Or did you find some non-determinism in the order of our files.
If so, totally fine to keep this, but feel free to open an issue, because we should be keeping the order intact.

ReeceHumphreys: We want the hash when you pass in foo.slice bar.slice to be the same as bar.slice foo.slice

PR #700, opened 2024-09-10
Improve Redefinition Validation

### LINKED ISSUE #699: Don't Check Child Elements when Parent is Redefined
If you write the same Slice definition twice, you'll get a redefinition error. For instance:
```
struct Foo { i: int32 }
struct Foo { i: int32 }
```
However, our redefinition check will continue to check inside the container.
So here. you'll get two redefinition errors, one for `Foo` and one for `i`.

But this 2nd error is somewhere between irrelevant or incorrect.

InsertCreativityHere:
This PR greatly improves the redefinition checking, part of which fixes #699.

The old check was very simple. We'd basically get every element's `scoped_identifier`, and check if any 2 were the same.
But this caused false positives, when some of the _parent_ scopes were redefined.

Now, we drill into each container, and check their contents separately.
So now the contents of one container has no influence on other containers.

Additionally, we also check things in the expected order. So we check a struct for redefinition before we check it's fields.
Before-hand, we just iterated through the nodes in whatever order the parser produced them (which doesn't usually line up with the 'expected' order).


src/parsers/common.rs
```diff
@@ -6,6 +6,7 @@ use crate::slice_file::Location;
 
 /// Stores a reference to a block of source code in a Slice file.
 #[derive(Clone, Copy, Debug)]
+#[allow(dead_code)] // The `end` field isn't used, but is very useful for debugging, so we keep it.
```
InsertCreativityHere, line 9: This field is only used by the `Debug` display implementation.

In the past, this counted as "being used", but a newer version of `rust` apparently doesn't count this anymore.
It's proven useful when debugging errors in the past at the cost of storing 2 extra `usize` per source-block.

PR #701, opened 2024-09-10
Simplify Slice File Hashing Logic

InsertCreativityHere:
This PR simplifies the logic used to compute hashes of our Slice files.
No more extension trait, with multiple implementations; just one function.

The only change in logic is that previously, we would compute a hash for each file separately, then hash all those hashes together:
```
total_hash = hash(hash(file1), hash(file2), hash(file3), ...)  // old
```
Now, we just have a single hash engine, which we directly feed all the files into:
```
total_hash = hash(file1, file2, file3, ...)  // new
```
I can't think of any adverse effects to this simpler approach, compared to the previous nested-hashing approach.

### DIFF
--- a/src/slice_file.rs
+++ b/src/slice_file.rs
@@ -4,7 +4,6 @@ use crate::grammar::*;
 use crate::utils::ptr_util::WeakPtr;
 use console::style;
 use serde::Serialize;
-use sha2::{Digest, Sha256};
 use std::cmp::{max, min, Ordering};
 use std::fmt::{Display, Write};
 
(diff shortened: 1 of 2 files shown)


src/slice_file.rs
```diff
@@ -167,53 +166,26 @@ impl SliceFile {
...
-    fn compute_sha256_hash_as_bytes(&self) -> [u8; 32] {
-        Sha256::new()
-            .chain_update(self.filename.as_bytes())
-            .chain_update(self.raw_text.as_bytes())
-            .finalize()
-            .into()
-    }
-}
+    // Filter out any reference files, and sort the source files which remain.
+    let mut sorted_sources: Vec<&SliceFile> = files.iter().filter(|f| f.is_source).collect();
```
externl, line 175: I forget where were landed on the discussion of only including src files. Was this previously being done elsewhere?

externl (approved):

src/slice_file.rs
InsertCreativityHere: Yes, you're right I should of added a comment mentioning this.
Currently, we do the filtering in `slicec-cs`:
```
let src_files: Vec<&SliceFile> = compilation_state.files.iter().filter(|f| f.is_source).collect();
let hash = SliceFile::compute_sha256_hash(&src_files);
```
https://github.com/icerpc/icerpc-csharp/blob/9edecc0ccaddee829828f2da64ce150b692d340f/tools/slicec-cs/src/main.rs#L65-L67
My thinking was to move it in `slicec`, to keep all the hashing logic and what/how we do it in a single place.

----

I'm also not completely sold on one vs the other, but this is the current behavior we have, so I kept it.
I'll blog about it, and see if we can reach a conclusion.

pepone (approved):
Much simpler, nice!

PR #704, opened 2025-11-28
Version 0.3.3 - Update Dependencies, CHANGELOG and Bump MSRV to 1.82

InsertCreativityHere:
This PR bumps our crate version to `0.3.3` (up from `0.3.2`).

It updates all our dependencies to their latest versions.
I read through every CHANGELOG entry for all of them and updated our code accordingly.

Because of the new dependencies, our MSRV (the minimum Rust version you need to compile this create) has increased from `1.70` to `1.82`.

There are also some very minor improvements to error messages we emit and code structure (usually recommended by Clippy).

----

**NOTE:** There are still some newer Rust features / stabilized APIs we could look into using now that we've jumped from `1.70` to `1.82`. But this is a larger effort than simply updating out dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,8 +17,7 @@ exclude = [".github/*", ".vscode/*"]
 
 [dependencies]
 # derive feature allows structs to derive Parser automatically
-# We lock the minor version to ensure compatibility with compilers that pass options to this library.
-clap = { version = "~4.5.53", features = ["derive"] }
+clap = { version = "4.5.53", features = ["derive"] }
 console = "0.16.1"
 convert_case = "0.10.0"
 in_definite = "1.1.2"


.vscode/extensions.json
```diff
@@ -1,7 +1,6 @@
 {
     "recommendations": [
         "rust-lang.rust-analyzer",
-        "streetsidesoftware.code-spell-checker",
-        "ZeroCInc.slice"
```
InsertCreativityHere: The Slice extension is useless for this Rust repository,
even if the repository is all about Slice.

src/ast/node.rs
```diff
@@ -25,8 +25,8 @@ macro_rules! generate_try_from_node_impl {
                     Ok($convert(x))
                 } else {
                     Err(LookupError::TypeMismatch {
-                        expected: stringify!($variant).to_case(Case::Lower),
-                        actual: node.to_string().to_case(Case::Lower),
+                        expected: ccase!(lower, stringify!($variant)),
+                        actual: ccase!(lower, node.to_string()),
```
InsertCreativityHere: `convert_case` now recommends using their new macro call instead of explicit function calls.

It seems a little more magical... but it is shorter (code-wise) and the recommended syntax, so, I updated to use it.

src/grammar/elements/operation.rs
```diff
@@ -88,19 +88,19 @@ impl Operation {
     }
 
     pub fn compress_arguments(&self) -> bool {
-        self.find_attribute::<Compress>().map_or(false, |a| a.compress_args)
+        self.find_attribute::<Compress>().is_some_and(|a| a.compress_args)
```
InsertCreativityHere: New convenience function for mapping `Option<T>`. The logic is identical.
If it's `None`, we get `false`, otherwise we get the result of the lambda.

Same as before where if it was `None`, we would get a default of `false`.

src/parsers/comments/mod.rs
```diff
@@ -52,7 +52,7 @@ fn construct_lint_from(parse_error: ParseError, file_name: &str) -> Diagnostic {
 fn generate_message(expected: &[String], found: impl std::fmt::Debug) -> String {
     let expected_message = match expected {
         [] => "expected no tokens".to_owned(),
-        [first] => format!("expected one of {first}"),
+        [first] => format!("expected {first}"),
```
InsertCreativityHere: Minor tweak to the error message we emit when we encounter a syntax error where there was only a single kind of token we expected.

It just felt weird to say "one of", then only list one thing.
**Old**: `expected one of "::", but found ")"`
**New**: `expected "::", but found ")"`

build.rs
```diff
@@ -4,5 +4,5 @@ extern crate lalrpop;
 
 fn main() {
     // Recursively finds any files ending with `.lalrpop` in the `src` directory and generates parsers from them.
```
InsertCreativityHere: `process_src` checks `./src` and is what 99% of projects should use.
`process_root` checks `.` and is for repos with multiple crates in them.

There was a documentation bug in LALRPOP where these two were flipped. The bug was fixed, and so is our build script.

Cargo.toml
```diff
@@ -12,24 +12,25 @@ readme = "README.md"
 keywords = ["slice", "ice", "icerpc"]
 license = "Apache-2.0"
 edition = "2021"
-rust-version = "1.70"
+rust-version = "1.82"
```
InsertCreativityHere: The biggest change is that `slicec` now requires a minimum of Rust 1.82 to use, because of our dependencies.

The current Rust version as of writing this is 1.91.

pepone (approved):

PR #706, opened 2026-01-15
Move `slicec` into it's own Directory

InsertCreativityHere:
This PR is the first part of converting the `slicec` repository into a Rust workspace (a repo with multiple crates in it).
This part isn't very interesting, it's just a matter of moving `slicec` into it's own folder named `slicec` (go figure).

In the next PR, I'll add the 2nd crate (`slice-codec`) which will be more interesting.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -2,7 +2,7 @@ name: coverage
 
 on:
   push:
-    branches: ["main", "phase-2"]
+    branches: ["main"]
 
 jobs:
   coverage:


slicec/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+name = "slicec"
+version = "0.3.3"
+authors = ["ZeroC Inc."]
+description = """
+The Slice parser and other core components for Slice compilers.
+"""
+homepage = "https://docs.icerpc.dev/slice2"
+documentation = "https://docs.rs/slicec"
+repository = "https://github.com/icerpc/slicec/slicec"
```
InsertCreativityHere: The only line that changed in this file is this one, where the path was updated to include the subfolder.

externl (approved):

PR #710, opened 2026-01-19
Added Preliminary Slice Definitions for `Ast`/`SliceFile`

InsertCreativityHere:
This PR adds an initial draft of Slice definitions to represent the AST (& the elements in it) and a SliceFile.

I thought about 2 approaches to encoding:
1) A `Sequence<Definition>` where each element is "thick" -- it's fully encapsulated and contains all the information you'd need for it.
2) A relational database, where the AST still has a `Sequence<Definition>`, but these nodes are tiny.
And instead, the AST had dictionaries for things like "which encodings does this support?", "where is this located?", etc.

In the end, I went with option 1. I think it is simplest to have the data in the node, instead of moved out into a 'database'. Maybe I'm too blinded by `slicec`'s design here.

----

There are 4 files:
- **Common.slice**: Defines the tiny building blocks that every file is going to want to use.
- **Structure.slice**: Defines the high level constructs like `SliceFile`, `Defintion`, and eventually the other things we want in the Request/Response.
- **SyntaxElements.slice**: Defines structs for each Slice element (struct, class, operation, Sequence, etc.).
- **DocComment.slice**: Defines all the doc-comment tags and structure.

## Open Question

Where should these files live? For `icerpc-slice` we have a separate repo, which we pull into each of the repos which need them (with `git-subtree`). Should we do the same for these with a `slicec-slice` repo? Or should we just copy/paste them into the repos that need them, since there's only 4 of these files, compared to the 21 in `icerpc-slice`.

### DIFF
--- a/slicec/slice/Structure.slice
+++ b/slicec/slice/Structure.slice
@@ -41,8 +41,9 @@ compact unchecked enum DefinitionKind { // TODO: we currently don't support 'unc
 struct Diagnostic {
     level: DiagnosticLevel,
     message: string,
-    span: Span?,
-    sliceId: SliceId?,
+    sliceId: SliceId?, // The SliceId of the thing that caused the diagnostic.
+    span: Span?, // Or, if no SliceId, the exact location of the diagnostic.
+    scope: SliceId?, // Needed for 'allow'/'warn' attributes on modules.
 }
 
 unchecked enum DiagnosticLevel : uint8 {
--- a/slicec/slice/SyntaxElements.slice
+++ b/slicec/slice/SyntaxElements.slice
@@ -19,7 +19,6 @@ struct Struct {
     isCompact: bool,
     attributes: Sequence<Attribute>,
     comment: DocComment?,
-    span: Span,
 }
 
 struct Class {
@@ -29,7 +28,6 @@ struct Class {
     base: TypeRef?,
     attributes: Sequence<Attribute>,
     comment: DocComment?,
-    span: Span,
 }
 
 struct Exception {
@@ -38,7 +36,6 @@ struct Exception {
     base: TypeRef?,
     attributes: Sequence<Attribute>,
     comment: DocComment?,
-    span: Span,
 }
 
 struct Field {
@@ -47,7 +44,6 @@ struct Field {
     \tag: Integer?,
     attributes: Sequence<Attribute>,
     comment: DocComment?,
-    span: Span,
 }
 
 struct Interface {
@@ -56,7 +52,6 @@ struct Interface {
     bases: Sequence<TypeRef>,
     attributes: Sequence<Attribute>,
     comment: DocComment?,
-    span: Span,
 }
 
 struct Operation {
@@ -68,7 +63,6 @@ struct Operation {
     encoding: Encoding,
     attributes: Sequence<Attribute>,
     comment: DocComment?,
-    span: Span,
 }
 
 struct Parameter {
@@ -77,7 +71,6 @@ struct Parameter {
     \tag: Integer?,
     isStreamed: bool,
     attributes: Sequence<Attribute>,
-    span: Span,
 }
 
 struct Enum {
@@ -88,7 +81,6 @@ struct Enum {
     isUnchecked: bool,
     attributes: Sequence<Attribute>,
     comment: DocComment?,
-    span: Span,
 }
 
 struct Enumerator {
@@ -97,7 +89,6 @@ struct Enumerator {
     fields: Sequence<Field>?,
     attributes: Sequence<Attribute>,
     comment: DocComment?,
-    span: Span,
 }
 
 compact enum Discriminant {
@@ -109,7 +100,6 @@ struct CustomType {
     identifier: Identifier,
     attributes: Sequence<Attribute>,
     comment: DocComment?,
-    span: Span,
 }
 
 struct TypeAlias {
@@ -117,7 +107,6 @@ struct TypeAlias {
     underlying: TypeRef,
     attributes: Sequence<Attribute>,
     comment: DocComment?,
-    span: Span,
 }
 
 struct ResultType {
@@ -154,16 +143,11 @@ compact enum Primitive : uint8 {
     \AnyClass,
 }
 
-struct Attribute {
-    directive: string,
-    args: Sequence<string>,
-}
-
 struct TypeRef {
     value: TypeRefDefinition,
     isOptional: bool,
     attributes: Sequence<Attribute>, /* We'll need to duplicate the attributes during encoding. */
-    span: Span,
+    span: Span, // Useful to report bad types (i.e. A Slice1 type is used and this is the Rust plugin).
 }
 
 enum TypeRefDefinition {
@@ -172,4 +156,15 @@ enum TypeRefDefinition {
     SequenceType(v: SequenceType),
     DictionaryType(v: DictionaryType),
     ResultType(v: ResultType),
+}
+
+struct Attribute {
+    directive: string,
+    args: Sequence<AttributeArgument>,
+    span: Span, // Useful to report bad attributes.
+}
+
+struct AttributeArgument {
+    value: string,
+    span: Span, // Useful to report bad arguments.
 }


slicec/slice/structure.slice
```diff
@@ -0,0 +1,39 @@
...
+    compilationMode: FileCompilationMode?,
+    moduleDeclaration: AstPointer?,
+    attributes: Sequence<AstPointer>,
+    contents: Sequence<AstPointer>,
+
+    isSource: bool,
+}
+
+struct Ast {
+    elements: Sequence<Node>,
```
InsertCreativityHere: ## Open Question

In `slicec`, the AST also has lookup-tables for finding elements by name.
Should we encode these? I lean towards no because I don't think consumers will be looking up elements by name.

slicec/slice/common.slice
```diff
@@ -0,0 +1,32 @@
...
+compact struct Location {
+    row: varuint62,
+    col: varuint62,
+}
+
+compact struct Span {
+    start: Location,
+    end: Location,
+    file: string,
+}
```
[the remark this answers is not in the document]
externl: I think for it'll probably be on a case by case basis. Overall I'm inclined to just keep them.

slicec/slice/common.slice
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) ZeroC, Inc.
+
+module SliceC
```
bernardnormier: Pretty clear to me all the definitions should be in a single file.

slicec/slice/Structure.slice
```diff
@@ -0,0 +1,29 @@
...
+}
+
+unchecked compact enum DefinitionKind { // TODO: we currently don't support 'unchecked compact'
+    Struct(v: Struct) = 0,
+    Class(v: Class) = 1,
+    Exception(v: Exception) = 2,
+    Interface(v: Interface) = 3,
+    Enum(v: Enum) = 4,
+    CustomType(v: CustomType) = 5,
+    TypeAlias(v: TypeAlias) = 6,
```
InsertCreativityHere: Note that these two things are effectively equivalent in Slice:
```
Interface (v: Interface),
Interface ({all the fields of `Interface`}),
```
Personally, I think it's cleaner/more readable to have actual structs for each definition type, instead of inlining all their fields directly in each enumerator.

And if the plugin really wants, it can decode as if the fields are directly in the enumerator anyways.

slicec/slice/Common.slice
```diff
@@ -0,0 +1,31 @@
...
+}
+
+compact struct Integer {
+    value: int128, // TODO: slice doesn't support int128
+    span: Span,
+}
+
+compact struct Identifier {
+    value: string,
+    span: Span,
```
bernardnormier: I find this span with every single identifier over the top. What's the use-case?

slicec/slice/SyntaxElements.slice
```diff
@@ -0,0 +1,175 @@
...
+    isCompact: bool,
+    isUnchecked: bool,
+    attributes: Sequence<Attribute>,
+    comment: DocComment?,
+    span: Span,
+}
+
+struct Enumerator {
+    identifier: Identifier,
+    value: EnumeratorValue,
```
bernardnormier: The correct term is discriminant:
https://docs.icerpc.dev/slice2/language-guide/enum-types#enumerators---with-fields

slicec/slice/SyntaxElements.slice
```diff
@@ -0,0 +1,175 @@
...
+    fields: Sequence<Field>?,
+    attributes: Sequence<Attribute>,
+    comment: DocComment?,
+    span: Span,
+}
+
+compact enum EnumeratorValue {
+    Implicit(v: int128), // TODO slice doesn't support int128,
```
bernardnormier: Why do we want to expose this implicit vs explicit discriminant value to the consumer?

bernardnormier (commented):
Would be nice to add doc-comments to these definitions.

slicec/slice/Common.slice
```diff
@@ -0,0 +1,32 @@
...
+compact struct Location {
+    row: varuint62,
+    col: varuint62,
+}
+
+compact struct Span {
+    start: Location,
+    end: Location,
+    file: string,
+}
```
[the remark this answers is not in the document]
bernardnormier: I think we should remove them for most definitions. Maybe keep them only for attributes, or attribute components.

InsertCreativityHere:
> Would be nice to add doc-comments to these definitions.

Yes, that's what I plan to do next, but I wanted to make sure everyone was at least mostly okay with the structure before I spent any time writing up nice doc-comments for them!

slicec/slice/Common.slice
```diff
@@ -0,0 +1,31 @@
...
+    value: string,
+    span: Span,
+}
+
+compact enum Encoding : uint8 {
+    Slice1 = 1,
+    Slice2 = 2,
+}
+
+typealias SliceId = string
```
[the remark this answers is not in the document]
InsertCreativityHere: It's the extension of what we currently call `TypeId` to also include such things as exceptions, operations, enumerators, etc. Every Slice element has an Id, hence `SliceId`. Some of those elements happen to be types, and in these cases, their `SliceId` is also a `TypeId`.

As always, open to other names! But we should really come up with a term for this instead of saying "TypeId but also for exceptions, etc."

slicec/slice/Common.slice
```diff
@@ -0,0 +1,32 @@
...
+}
+
+compact struct Integer {
+    value: int128, // TODO: slice doesn't support int128
+    span: Span,
+}
+
+compact struct Identifier {
+    value: string,
+    span: Span,
```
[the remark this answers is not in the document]
InsertCreativityHere: Indeed. This is why me and externl yesterday were trying to find an alternative way to reference a Slice element... I came up with an addressing scheme, but it was pretty janky for pointing to things like "the 3rd argument to the 2nd attribute on my struct". So I stuck with the spans.

After all of us talking it over, I think a hybrid approach is best. For things with a `SliceId` (fully-scoped-identifier), we can completely remove their spans. We can just return their ID, and let `slicec` lookup the location.

slicec/slice/DocComment.slice
```diff
@@ -0,0 +1,50 @@
...
+    params: Sequence<ParamTag>,
+    returns: Sequence<ReturnsTag>,
+    \throws: Sequence<ThrowsTag>,
+    see: Sequence<SeeTag>,
+    span: Span,
+}
+
+struct ParamTag {
+    identifier: Identifier,
+    message: Message,
```
externl: How to empty messages work? If we allow them and this is fine then should DocComment::overview be non-optional?

slicec/slice/SyntaxElements.slice
```diff
@@ -154,16 +143,11 @@ compact enum Primitive : uint8 {
...
-    directive: string,
-    args: Sequence<string>,
-}
-
 struct TypeRef {
     value: TypeRefDefinition,
     isOptional: bool,
     attributes: Sequence<Attribute>, /* We'll need to duplicate the attributes during encoding. */
-    span: Span,
+    span: Span, // Useful to report bad types (i.e. A Slice1 type is used and this is the Rust plugin).
```
pepone: Doesn't this require the file using `Slice1` mode?

slicec/slice/Structure.slice
```diff
@@ -0,0 +1,52 @@
...
+
+struct InputFile {
+    relativePath: string,
+
+    compilationMode: FileCompilationMode?,
+    moduleDeclaration: Module?,
+    attributes: Sequence<Attribute>,
+    contents: Sequence<Definition>,
+
+    isSource: bool,
```
externl: Could we (read: are you not opposed to) make this an enum :) ?

slicec/slice/Structure.slice
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) ZeroC, Inc.
+
+module Slice::Compiler
+
+interface CodeGenerator {
+    generateCode(inputFiles: Sequence<InputFile>)
```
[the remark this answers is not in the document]
bernardnormier: A parameter list in Slice is comparable to a structure.

InsertCreativityHere:
I am closing this PR and opening a new one in it's place, since the definitions have evolved quite alot since my original commits!

PR #712, opened 2026-01-22
Minimal `codeGenerator` Definitions in Slice

InsertCreativityHere:
This PR shows the absolute smallest I think we can pack the definitions needed for `generateCode` down to.
In the end, only 19 types are necessary (and 2 of those are simple enums). As opposed to #711 which has 35 types.

Granted, sending the data modeled as a relational database actually gets us even fewer types (only 8), since much of the necessary information can be stored in dictionaries instead of needing all these break-out types with specialized fields.


InsertCreativityHere:
Nah.

PR #713, opened 2026-01-31
Change namespace and generateCode parameters

bernardnormier:
This PR makes small changes to the new Compiler definitions.

### DIFF
--- a/slicec/slice/CodeGenerator.slice
+++ b/slicec/slice/CodeGenerator.slice
@@ -1,10 +1,11 @@
 // Copyright (c) ZeroC, Inc.
 
-module Slice::Compiler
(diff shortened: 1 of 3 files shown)


slicec/slice/CodeGenerator.slice
```diff
@@ -14,7 +15,6 @@ struct SliceFile {
     tag(0) attributes: Sequence<Attribute>?
     contents: Sequence<Definition>
     anonymousTypes: Sequence<AnonymousType>
-    isSource: bool
```
bernardnormier: Being a source file or a reference file is not a property of the file. It's about how it's used in a specific situation.

Issue #724, opened 2024-05-06
Remove support for multiple slice encodings


ReeceHumphreys:
> We discussed this support for multiple Slice encodings a while ago and I made it clear that I don't want support for multiple versions of the Slice encoding in Rust.
> 
> There should be no reference in Slice2 or Slice1 in this code. The Slice encoder / encoder implements the "Slice encoding" (no explicit version), just like Protobuf doesn't have an explicit encoding version (because it never changed).

_Originally posted by @bernardnormier in https://github.com/icerpc/slice-rust/pull/11#pullrequestreview-2039048544_

PR #725, opened 2026-02-05
Removed Stubbed-out Code for Slice1 Encoding in Rust

InsertCreativityHere:
This PR removes all mentions of the `Slice1` encoding from the Rust `slice-codec`.
We never actually implemented this encoding, so there isn't alot to remove.

This implements #724.


bernardnormier (approved):
We should also remove all mentions of slice2 in a follow-up PR.

As far as the Rust Slice Codec is concerned, there is only one Slice encoding.

PR #729, opened 2026-03-06
Refactor the `slicec` Binary to Write Encoded Slice Definitions to 'stdout'

InsertCreativityHere:
This PR does what the title says! I haven't implemented the process forking (i.e. where you can pass code-gen plugins on the command-line and `slicec` will fork them). That's next up on the list.

But you can still pipe the output of `cargo run -- Test.slice` to another process, or a file.
It's obviously hard to test this without a decoder. I've been using `cargo run -- Test.slice > slice-output.dat`,
and opening the file in a hex editor to manually check the bytes. The bytes seem correct for normal Slice files.

----

### Structure

This PR does 3 big things, in 2 files:
- `definition_types.rs`: this file is my 'hand-mapping' the [Slice Compiler definitions](https://github.com/icerpc/icerpc-slice/tree/main/Compiler).
It will eventually be replaced by generated code when we add code-gen in Rust.
This file **1)** Defines the types that will be sent on-the-wire and **2)** implements the encoding of these types.
The encoding is complete, except for `DocComments` which require tags (which require size/position tracking which doesn't exist yet).
- `slice_file_converter.rs` is the more interesting one. This file provides conversion functions for mapping between domains. It takes types from `slicec` and converts them to the Slice-generated types. Most of this is straightforward, but there are some differences between them which required attention.

### DIFF
--- a/slicec/src/slice_file_converter.rs
+++ b/slicec/src/slice_file_converter.rs
@@ -177,6 +177,11 @@ impl From<&GrammarMessageComponent> for MessageComponent {
 
 /// This struct exposes a function ([`SliceFileContentsConverter::convert`]) that converts the contents of a Slice file
 /// from their AST representation, to a representation that can be encoded with the Slice encoding.
+//
+// This struct is necessary due to anonymous types, which need their own symbols. So, when you convert a `Field`, that
+// may need just a `Field` symbol, but it might also need a `Field`, `Sequence`, and `Dictionary` symbol if the field's
+// type uses a sequence of dictionaries. To handle this, we need to keep some state (`converted_contents`), which
+// symbols can be pushed into at any time. Since there's no way to know how many symbols a definition will need upfront.
 #[derive(Debug)]
 pub struct SliceFileContentsConverter {
     converted_contents: Vec<Symbol>,


slice-codec/src/slice2/encoding.rs
```diff
@@ -151,6 +156,13 @@ impl EncodeInto<Slice2> for &str {
     }
 }
 
+#[cfg(feature = "alloc")]
+impl EncodeInto<Slice2> for &String {
+    fn encode_into(self, encoder: &mut Encoder<impl OutputTarget, Slice2>) -> Result<()> {
+        self.as_str().encode_into(encoder)
```
InsertCreativityHere: The encoding of `&String` just delegates to the encoding for `&str`.

This is purely to save the caller from needing to do the conversion themselves.

slicec/src/definition_types.rs
```diff
@@ -0,0 +1,328 @@
+// Copyright (c) ZeroC, Inc.
```
InsertCreativityHere: In this file, the most important thing to scrutinize is that the encoding logic is correct.

slicec/src/slice_file_converter.rs
```diff
@@ -0,0 +1,358 @@
+// Copyright (c) ZeroC, Inc.
+
+// Pull in the core compiler types using aliases to disambiguate them from the Slice-compiler definitions.
+// Any type that starts with 'Compiler' is a compiler type, not a mapped Slice-compiler definition type.
```
InsertCreativityHere: Not sure if there's a better solution here... Alot of type names are identical between `slicec` and the `Compiler Definitions`; unsurprisingly.

slicec/src/slice_file_converter.rs
```diff
@@ -0,0 +1,358 @@
...
+#![cfg_attr(rustfmt, rustfmt_skip)] // skip the `use ... as ...` to keep them one-per-line.
```
InsertCreativityHere: `rustfmt` does not handle aliased imports well apparently. It tries to format like normal imports, which makes them unreadable.
Tbf, aliases are pretty uncommon in practice.

slicec/src/slice_file_converter.rs
```diff
@@ -0,0 +1,358 @@
...
+    fn from(component: &CompilerMessageComponent) -> Self {
+        match component {
+            CompilerMessageComponent::Text(text) => MessageComponent::Text(text.clone()),
+            CompilerMessageComponent::Link(tag) => {
+                MessageComponent::Link(convert_doc_comment_link(tag.linked_entity()))
+            }
+        }
+    }
+}
+
```
InsertCreativityHere: For everything else, we need to keep some state. So we implement their conversion functions inside a struct which can hold that state.

slicec/src/main.rs
```diff
@@ -16,13 +57,37 @@ fn main() {
...
-        println!("{diagnostic:?}");
+    // TODO: replace this by forking a code-gen plugin once they exist.
+    // For now, if there are any diagnostics, we emit those and NOT the encoded definitions.
+    // Code-generators can tell if it's okay to decode or not by the presence of the `"generateCode"` string.
+    if totals.0 + totals.1 > 0 {
+        // If there were diagnostics, print them to 'stdout' and don't encode the Slice definitions.
+        print!("Diagnostics: ");
+        println!("{totals:?}");
+        for diagnostic in updated_diagnostics {
+            println!("{diagnostic:?}");
```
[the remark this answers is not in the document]
InsertCreativityHere: This code is all temporary, and will be replaced with forking soon.

But these diagnostics do not indicate errors in the program itself, they are only errors in the user's inputs. These are different things. Even if we hit this case, the program is functioning correctly.

slicec/src/definition_types.rs
```diff
@@ -0,0 +1,328 @@
...
+}
+impl EncodeInto<Slice2> for &MessageComponent {
+    fn encode_into(self, encoder: &mut Encoder<impl OutputTarget>) -> Result<()> {
+        // Write the discriminant value.
+        // SAFETY: this cast is guaranteed to be safe because the enum is marked with `repr(u8)`, so it's safe to cast
+        // it directly to a `u8`.
+        unsafe {
+            let discriminant = *<*const _>::from(self).cast::<u8>();
+            encoder.encode_varint(discriminant)?;
+        }
```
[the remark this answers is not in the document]
InsertCreativityHere: Read my response to the first comment generated on this PR.
Copilot is just wrong, and this is a guarantee the language makes and not UB.
To my knowledge, it's misunderstanding Rust here.

slicec/src/slice_file_converter.rs
```diff
@@ -0,0 +1,359 @@
...
+            failure_type: self.convert_type_ref(&result_type.failure_type),
+        }
+    }
+
+    /// Returns a [TypeId] for the provided `type_ref`. This is a fully-scoped identifier for user-defined types,
+    /// the corresponding keyword for primitive types, and for anonymous types, we do the following:
+    /// 1) Recursively convert the anonymous type (and any nested types) to the mapped definition types.
+    /// 2) Add these directly to [Self::converted_contents] (so these types appear in the contents before their users)
+    /// 3) Return its index in [Self::converted_contents] as a numeric TypeId.
+    fn get_type_id_for(&mut self, type_ref: &CompilerTypeRef) -> TypeId {
```
bernardnormier: We should find a better name for this method. You don't expect a get_xxx method to modify self.

PR #746, opened 2026-03-26
`slicec` Now Decodes Responses from Code-Generation Plugins

InsertCreativityHere:
This PR changes `slicec` to fork plugins as subprocesses, using the standard streams to communicate with them.
It also adds the logic for `slicec` to receive responses from these code-generator plugins, which are decoded, and then used to write any generated files to disk.

If you want to try it yourself just run: `cargo run -- <arguments to slicec>`.
But the code-generators have to be in your path to run. If you don't, `slicec` will emit:
```
error [E001]: unable to run code-generator plugin 'IceRpc.Slice.Generator.exe': No such file or directory
```

----

The CLI will be added in a follow-up PR, wanted to keep each PR focused. This one is about the plugin forking and communication, next will be about completely refactoring the `slicec` command line options.

### DIFF
--- a/slicec/src/main.rs
+++ b/slicec/src/main.rs
@@ -2,7 +2,7 @@
 
 use std::fs::File;
 use std::io::{Error, ErrorKind, Write};
-use std::process::{Child, Command, Stdio};
+use std::process::{Child, Command, ExitCode, Stdio};
 
 use clap::Parser;
 
@@ -132,8 +132,19 @@ fn handle_plugin_response(response_payload: Vec<u8>) -> std::io::Result<Diagnost
 }
 
 fn write_generated_file(generated_file: &definition_types::GeneratedFile) -> std::io::Result<()> {
+    let generated_file_bytes = generated_file.contents.as_bytes();
+
+    // If the generated file already exists on disk, and is identical to what we want to write,
+    // we don't overwrite the file, and instead return immediately.
+    if let Ok(current_contents) = std::fs::read(&generated_file.path) {
+        if current_contents == generated_file_bytes {
+            return Ok(());
+        }
+    }
+
+    // Write the generated file to disk.
     let mut file = File::create(&generated_file.path)?;
-    file.write_all(generated_file.contents.as_bytes())?;
+    file.write_all(generated_file_bytes)?;
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


slicec/src/main.rs
```diff
@@ -46,50 +50,160 @@ fn encode_generate_code_request(parsed_files: &[slicec::slice_file::SliceFile])
...
+        mut diagnostics,
+        files,
+    } = compilation_state;
+
+    // Only invoke the plugins if there were no errors in the Slice files.
+    if !diagnostics.has_errors() {
+        // Encode the request which will be sent to each of the code-generation plugins.
+        let encoded_request = encode_generate_code_request(&files)?;
+
+        // TODO: add a CLI to choose which plugins are run, instead of hard-coding them here.
```
bernardnormier: You mean add `--generator=...` command-line option to slicec as we discussed?

slice-codec/src/error.rs
```diff
@@ -26,7 +29,7 @@ pub struct Error {
 
     /// The underlying cause of this error, if any exist.
     #[cfg(feature = "alloc")]
-    source: Option<Box<dyn core::error::Error + 'static>>,
+    source: Option<Box<dyn core::error::Error + Send + Sync + 'static>>,
```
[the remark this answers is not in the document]
bernardnormier: Are we doing any multithreading in slicec?

slicec/src/main.rs
```diff
@@ -46,50 +50,160 @@ fn encode_generate_code_request(parsed_files: &[slicec::slice_file::SliceFile])
...
+    let mut slice_decoder = Decoder::from(&response_payload);
+    let generated_files: Vec<definition_types::GeneratedFile> = slice_decoder.decode()?;
+    let plugin_diagnostics: Vec<definition_types::Diagnostic> = slice_decoder.decode()?;
+
+    // TODO: Convert the diagnostics we decode from the plugin, into diagnostics that slicec can handle.
+    //       To do this requires re-working the diagnostic API fairly substantially.
+    for plugin_diagnostic in plugin_diagnostics {
+        println!("{}", plugin_diagnostic.message);
+    }
+    let mut diagnostics = Diagnostics::new();
```
InsertCreativityHere: There is no magic happening here. This is just a dummy variable for now.
In a future PR, the `for` loop will map diagnostics into a form  `slicec` can handle, but right now it just `println`s them.

pepone (approved):

bernardnormier (approved):

PR #751, opened 2026-03-30
Remove compilation mode and encoding support

externl:
- Delete compilation mode, encoding, tag format, and supported
  encodings types and logic from grammar
- Remove encoding patcher and supported encodings module
- Remove `mode` keyword from lexer, parser, and token definitions
- Remove compilation-mode-related error variants and renumber
  error codes E024–E036
- Drop `supported_encodings` and `encoding` fields from grammar
  elements (Struct, Enum, Interface, CustomType, TypeAlias,
  Operation, etc.)
- Remove `tag_format()` and `supported_encodings()` from Type
  trait and all implementors
- Delete compilation mode, mixed encoding, mode compatibility,
  and Slice1-specific tests
- Simplify remaining tests to remove `mode = SliceN` usage

### DIFF
--- a/slicec/tests/typealias_tests.rs
+++ b/slicec/tests/typealias_tests.rs
@@ -10,14 +10,14 @@ mod typealias {
     use slicec::slice_file::Span;
     use test_case::test_case;
 
-    #[test_case("struct S {}", "S" ; "structs")]
-    #[test_case("enum E { Foo }", "E" ; "enums")]
-    #[test_case("custom C", "C" ; "custom types")]
-    #[test_case("", "bool" ; "primitives")]
-    #[test_case("", "Sequence<bool>" ; "sequences")]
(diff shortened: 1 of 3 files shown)


slicec/tests/typealias_tests.rs
```diff
@@ -138,38 +137,11 @@ mod typealias {
         check_diagnostics(diagnostics, [expected]);
     }
 
-    #[test_case("Slice1", "uint32"; "Slice1")]
-    fn reject_underlying_types_based_on_mode(mode: &str, underlying_type: &str) {
+    #[test_case("uint32"; "uint32")]
```
InsertCreativityHere: I'd just delete this test. We already test underlying types in a test near the top of this file.
This was just testing that certain types are allowed or disallowed depending on compilation mode, which no longer exists.

PR #752, opened 2026-03-30
Add Command-Line Options for Specifying Generators

InsertCreativityHere:
This PR adds support for a new `--generator` (or `-G`) command-line option, which allows users to specify which generators should be invoked by `slicec` after parsing and validation (only if those succeed of course).

Here is the help message describing the new option:
```
  -G, --generator <GENERATOR>
          Specify a code-generator plugin that should be run after parsing and validation complete (if successful).
            Ex: '--generator /path/to/my/plugin'
          
          Each code-generator can be provided with arbitrary string arguments using the following syntax:
            '/path/to/my/plugin;arg1=value1;arg2 = value2;arg3;...'
          Leading and trailing whitespace is stripped from arguments and their values. Argument values are optional.
```

### DIFF
--- a/slicec/src/definition_types.rs
+++ b/slicec/src/definition_types.rs
@@ -243,7 +243,7 @@ impl EncodeInto for &MessageComponent {
 }
 
 pub struct Arguments(pub Vec<(String, String)>);
-impl EncodeInto<Slice2> for Arguments {
+impl EncodeInto for Arguments {
     fn encode_into(self, encoder: &mut Encoder<impl OutputTarget>) -> Result<()> {
         encoder.encode_size(self.0.len())?;
         for (e1, e2) in &self.0 {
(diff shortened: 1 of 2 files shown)


slicec/src/definition_types.rs
```diff
@@ -242,6 +242,18 @@ impl EncodeInto<Slice2> for &MessageComponent {
     }
 }
 
+pub struct Arguments(pub Vec<(String, String)>);
```
InsertCreativityHere: There's two common 'dictionary' representations in Rust:
- `HashMap` if you don't care about ordering
- `Vec<(String, String)>` if you care about ordering

Here, it'd be nice to preserve the order of arguments, so we use the latter.

pepone (approved):

PR #756, opened 2026-04-01
Removed More `slicec` Code that is no Longer Needed

InsertCreativityHere:


slicec/tests/sequence_tests.rs
```diff
@@ -21,14 +21,14 @@ mod sequences {
 
         // Assert
         let seq_def = ast.find_element::<TypeAlias>("Test::Seq").unwrap();
```
InsertCreativityHere: The one test that was using `concrete_typeref` and `TypeRefs`.

Reworked to use `concrete_type` and `Types` instead.
No idea why it wasn't using those to begin with.

slicec/tests/test_helpers.rs
```diff
@@ -55,10 +51,12 @@ pub fn parse_for_diagnostics(slice: impl Into<String>) -> Vec<Diagnostic> {
 /// Each string is treated as a separate Slice file by the parser.
 #[must_use]
 pub fn parse_multiple_for_diagnostics(slice: &[&str]) -> Vec<Diagnostic> {
```
InsertCreativityHere: `diagnostics_from_compilation_state` was defined in the `src/test_helpers.rs`.
This is the only place it was used, so I just inlined the logic.

Issue #764, opened 2026-04-08
Move 'attribute_parsing_util'


InsertCreativityHere:

InsertCreativityHere:
This was implemented in #766.

PR #766, opened 2026-04-09
Refactor Attribute Diagnostics

InsertCreativityHere:
This PR refactors how `slicec` reports attribute errors, ahead of me adding a better API for generators to report attribute errors.

1) Consolidated and renamed the possible attribute errors (and added doc-comments to them).
2) Removed 'attribute_parsing_util.rs' (#764).
Replaced it's 4 functions, with only 2 (now in the `slicec::grammar::attributes` module)

The errors I've defined in `errors.rs` are identical to the ones that `slicec` will accept over the wire from Code-Generators (in my next PR), so review them with this context in mind.


slicec/src/grammar/attributes/deprecated.rs
```diff
@@ -11,7 +11,7 @@ impl Deprecated {
     pub fn parse_from(Unparsed { directive, args }: &Unparsed, span: &Span, diagnostics: &mut Diagnostics) -> Self {
         debug_assert_eq!(directive, Self::directive());
 
-        check_that_at_most_one_argument_was_provided(args, Self::directive(), span, diagnostics);
+        check_argument_count_is_within(0..2, args, Self::directive(), span, diagnostics);
```
InsertCreativityHere, line 14: Note that ranges in rust are "half-open" ranges.

In mathematical notation: `0..2` = `[0,2)`, so this range only includes `0`, and `1`.
In English: Rust ranges include the starting value, but exclude the ending value.

----

You don't need to read this:

But Austin, why not use the alternate syntax of `0..=1` instead of `0..2`?
Which makes it easier to see that this accepts 0 or 1 arguments!

These correspond to different constructs in Rust,
`x..y` is a shorthand for `Range`, and `x..=y` is a shorthand for `RangeInclusive`.
TL;DR, `RangeInclusive` is actually a pain to use, and has way less functionality than `Range`, all because it's possible to type `0..=u8::MAX`. The length of this range is 256, which does not fit in a `u8`!
Unlike `0..u8::Max` which only has a length of 255 (since the end value isn't included).

The fact that `RangeInclusive` can create ranges larger than their backing type means it can't safely implement alot of the same functions as `Range`. Which in turn makes it less useful.

slicec/src/grammar/attributes/mod.rs
```diff
@@ -75,3 +76,49 @@ impl AttributeKind for Unparsed {
         &self.directive
     }
 }
+
```
InsertCreativityHere, line 79: Helper functions.

These are spiritual successors to the ones deleted from `attribute_parsing_util`, and should be reviewed.

slicec/src/grammar/attributes/mod.rs
```diff
@@ -75,3 +76,49 @@ impl AttributeKind for Unparsed {
...
+    if let Some(note) = note {
+        diagnostic = diagnostic.add_note(note, None);
+    }
+
+    diagnostic.push_into(diagnostics);
+}
+
+/// Reports an error if an incorrect number of attributes was provided to the specified attribute.
+///
+/// # Arguments
```
InsertCreativityHere, line 101: This is the Rusty syntax for documenting function arguments. But in Rust, you should only document arguments if there's alot or if it's not clear what they do just from the summary and it's name.

So, it's actually rare that you need to document your arguments like this.

slicec/src/diagnostics/errors.rs
```diff
@@ -179,33 +181,52 @@ pub enum Error {
...
-        attribute: String,
+    /// Too few or too many arguments were supplied to an otherwise valid attribute.
+    /// For example: `[oneway(FooBar)]` ('oneway' takes no arguments) or `[compress]` ('compress' requires arguments).
+    IncorrectAttributeArgumentCount {
+        /// The directive of the attribute.
+        directive: String,
+        /// A range representing the number of arguments this attribute can be given.
+        expected_count: Range<usize>,
+        /// The number of arguments this attribute was actually given.
+        actual_count: usize,
```
pepone, line 229: InvalidAttributeArgument, IncorrectAttributeArgumentCount

What is the point of distinguishing between the two?

slicec/src/diagnostics/errors.rs
```diff
@@ -342,93 +356,129 @@ implement_diagnostic_functions!(
...
-        format!("missing required attribute '{attribute}'"),
-        attribute
+        InvalidAttributeArgument,
+        format!("'{argument}' is not a valid argument to the '{directive}' attribute"),
+        argument,
+        directive
     ),
     (
         "E028",
+        IncorrectAttributeArgumentCount,
```
pepone, line 421: Seems quite complicated, IMO the invalid attribute argument is enough here.

pepone (approved):

externl (approved):

slicec/src/diagnostics/errors.rs
InsertCreativityHere: Because they just feel like different errors to me.
`[compress(FooBar)]` the error is that `FooBar` isn't a valid argument. It has to be either `Args` or `Returns`
`[oneway(FooBar)]` the error is that `oneway` shouldn't take any arguments at all.

IMO it's nice to have separate, more clearer error messages for the user instead of something more generic like:
"invalid attribute arguments supplied"
`slicec` has always tried to give very rich and useful errors, like `rust` does, and unlike our old compiler does.

slicec/src/diagnostics/errors.rs
InsertCreativityHere: The complexity is just from me trying to emit the most accurate and grammatically correct message.
If you really want, I can prune the grammar corrections, and not bother handling singular vs plural.
Personally, I don't think is so bad tho.

slicec/src/diagnostics/errors.rs
InsertCreativityHere: For what it's worth, with this scheme, the C# generator will never throw `InvalidArgumentException`, only `IncorrectAttributeArgumentCount`, since it never validates the inner argument.

PR #769, opened 2026-04-13
Slight Tweaks to Slice Compiler Definitions

InsertCreativityHere:
This PR renames 'arguments' to 'options' to match the terminology used by `icerpc-csharp`.
It also re-orders the anonymous types so that they are in the canonical order we always use:
    `Result`, `Sequence`, `Dictionary`

### DIFF
--- a/slicec/src/definition_types.rs
+++ b/slicec/src/definition_types.rs
@@ -189,6 +189,13 @@ pub struct TypeAlias {
 implement_encode_into_for_struct!(TypeAlias, entity_info, underlying_type);
 
 #[derive(Clone, Debug)]
+pub struct ResultType {
+    pub success_type: TypeRef,
+    pub failure_type: TypeRef,
+}
+implement_encode_into_for_struct!(ResultType, success_type, failure_type);
+
+#[derive(Clone, Debug)]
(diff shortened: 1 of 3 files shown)


externl (approved):

slicec/src/definition_types.rs
pepone, line 269: What is the logic for this sorting?

InsertCreativityHere: I think it's loosely based on "similarity" and "language importance".

So, all the anonymous types should be next to one another, since they're conceptually similar. Same for all the user defined types. Typealias goes at the bottom because it's not similar to anything else, and is also the least important, and least useful of Slice constructs.

Then within each grouping, it's kind of just what feels right I guess. `Sequence` feels like it should come before `Dictionary`. `Result` should come before either of them because it has more in common with the user-defined types than it has with the collection types.

But this PR was just to make them consistent with the order we settled on them for slicec. If you have any other elements you want to swap around, feel free to comment!

`Custom` could be closer to `TypeAlias`, `Enum` could move beneath `Struct`, etc.

pepone (approved):

PR #770, opened 2026-04-15
Refactor Diagnostic Creation

InsertCreativityHere:
This PR:
- Removes the `implement_diagnostic_functions` macro we used to implement functions for diagnostic types. Now we just implement the functions like normal.
- Adds more typed constructors to create diagnostics, in anticipation of adding a new `Info` diagnostic kind, which will be very different from the existing `Error` and `Lint`s.

None of the actual error messages or codes should be changed. Just moved the implementation from a macro to functions.

### DIFF
--- a/slicec/src/diagnostics/diagnostic.rs
+++ b/slicec/src/diagnostics/diagnostic.rs
@@ -18,11 +18,12 @@ pub struct Diagnostic {
 }
 
 impl Diagnostic {
-    pub fn new(kind: impl Into<DiagnosticKind>) -> Self {
-        let kind = kind.into();
+    /// Creates a new `Diagnostic` directly from a [`DiagnosticKind`].
+    /// The newly created `Diagnostic` has no `span`, `scope`, or `notes` set.
+    pub fn new(kind: DiagnosticKind) -> Self {
         let level = match &kind {
             DiagnosticKind::Error(_) => DiagnosticLevel::Error,
-            DiagnosticKind::Lint(lint) => lint.get_default_level(),
+            DiagnosticKind::Lint(lint) => lint.default_diagnostic_level(),
         };
 
         Diagnostic {
@@ -32,6 +33,18 @@ impl Diagnostic {
             scope: None,
             notes: Vec::new(),
         }
+    }
+
+    /// Creates a new error `Diagnostic` from the provided [`Error`].
+    /// The newly created `Diagnostic` has no `span`, `scope`, or `notes` set.
+    pub fn error(error: Error) -> Self {
+        Self::new(DiagnosticKind::Error(error))
+    }
+
+    /// Creates a new lint `Diagnostic` from the provided [`Lint`].
+    /// The newly created `Diagnostic` has no `span`, `scope`, or `notes` set.
+    pub fn lint(lint: Lint) -> Self {
+        Self::new(DiagnosticKind::Lint(lint))
     }
 
     /// Returns the message of this diagnostic.
@@ -45,8 +58,8 @@ impl Diagnostic {
     /// Returns this diagnostic's code. This is either the name of a lint or of the form `E###`.
     pub fn code(&self) -> &str {
         match &self.kind {
-            DiagnosticKind::Error(error) => error.code(),
-            DiagnosticKind::Lint(lint) => lint.code(),
+            DiagnosticKind::Error(error) => error.error_code(),
+            DiagnosticKind::Lint(lint) => lint.lint_name(),
         }
     }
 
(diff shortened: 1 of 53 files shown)


slicec/src/diagnostics/mod.rs
```diff
@@ -11,76 +8,12 @@ pub use diagnostic::*;
 pub use errors::Error;
 pub use lints::Lint;
 
+use crate::slice_file::Span;
+use serde::Serialize;
+
```
InsertCreativityHere, line 13: The conventional ordering for top-level Rust statements is `pub mod`, `mod`, `pub use`, `use`.
So I moved these `use` statements to where they should be.

slicec/src/diagnostics/mod.rs
```diff
@@ -11,76 +8,12 @@ pub use diagnostic::*;
...
 /// Stores additional information about a diagnostic.
 #[derive(Serialize, Debug, Clone)]
 pub struct Note {
     pub message: String,
     pub span: Option<Span>,
 }
-
-/// A macro that implements the `code` and `message` functions for [Lint] and [Error] enums.
-macro_rules! implement_diagnostic_functions {
-    (Lint, $(($kind:ident, $message:expr $(, $variant:ident)* )),*) => {
```
InsertCreativityHere, line 23: Probably the fanciest macro we had in `slicec`.

slicec/src/grammar/attributes/allow.rs
```diff
@@ -2,6 +2,16 @@
 
 use super::*;
 
+/// All the valid arguments that can be supplied to the `allow` metadata, or the `--allow` command line option.
+pub const ALLOWABLE_LINT_IDENTIFIERS: [&str; 6] = [
```
InsertCreativityHere, line 6: This used to be generated by the macro, now we just write it by hand.
It does live in a different module though, so had to update it's scoping where we use it.

slicec/src/grammar/attributes/mod.rs
```diff
@@ -13,7 +13,7 @@ pub use oneway::*;
 pub use sliced_format::*;
 
 use super::Attributables;
-use crate::diagnostics::{Diagnostic, Diagnostics, Error, Lint};
```
InsertCreativityHere, line 16: `Lint` was in scope to see the `ALLOWABLE_LINT_IDENTIFIERS`. Now that it lives in this module ancestry, no longer needed.

slicec/src/diagnostics/diagnostic.rs
```diff
@@ -105,18 +118,6 @@ pub enum DiagnosticKind {
...
-    fn from(error: Error) -> Self {
-        DiagnosticKind::Error(error)
-    }
-}
-
-impl From<Lint> for DiagnosticKind {
-    fn from(lint: Lint) -> Self {
-        DiagnosticKind::Lint(lint)
-    }
-}
```
InsertCreativityHere, line 118: These are no longer needed now that we have dedicated constructors for `error` and `lint`, instead of using `From/Into` casts.

slicec/src/parsers/mod.rs
```diff
@@ -2,16 +2,16 @@
...
-
 mod comments;
 mod common;
 mod preprocessor;
 mod slice;
 
+// We only export the parsers and keep all the other logic private.
+pub use self::comments::parser::CommentParser;
+pub use self::preprocessor::parser::Preprocessor;
+pub use self::slice::parser::Parser;
```
InsertCreativityHere, line 13: `pub use` should come after `mod` statements.

slicec/src/grammar/attributes/allow.rs
```diff
@@ -2,6 +2,16 @@
...
+    "All",
+    "DuplicateFile",
+    "Deprecated",
+    "MalformedDocComment",
+    "IncorrectDocComment",
+    "BrokenDocLink",
+];
```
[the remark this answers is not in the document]
InsertCreativityHere: It's disallowed for attributes, but not on the command-line.
So we still need to keep it in this list. The attribute validation code already handles this. Nothing changed.

slicec/src/diagnostics/lints.rs
```diff
@@ -46,26 +47,36 @@ impl Lint {
...
+                } else {
+                    format!("'{identifier}' is deprecated")
+                }
+            }
+
+            Self::MalformedDocComment { message } => format!("malformed doc comment: {message}"),
+
+            Self::IncorrectDocComment { message } => format!("incorrect doc comment: {message}"),
+
+            Self::BrokenDocLink { message } => format!("broken doc link: {message}"),
```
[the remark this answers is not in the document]
InsertCreativityHere: Fair point. The lints did actually have a message change. You're right, and my PR description is slightly inaccurate.

slicec/src/diagnostics/lints.rs
```diff
@@ -46,26 +47,36 @@ impl Lint {
...
+        }
```
InsertCreativityHere, line 80: This is the one place where the messages changed. I added short little prefixes before emitting the messages. Before we just emitted `{message}` verbatim. No prefix.

slicec/src/diagnostics/diagnostic.rs
```diff
@@ -34,6 +35,18 @@ impl Diagnostic {
         }
     }
 
+    /// Creates a new error `Diagnostic` from the provided [`Error`].
+    /// The newly created `Diagnostic` has no `span`, `scope`, or `notes` set.
+    pub fn error(error: Error) -> Self {
```
externl: Shouldn't these be something like `from_error` and `from_info`?

externl (approved):
Looks good to me!

slicec/src/diagnostics/diagnostic.rs
InsertCreativityHere: Yes you're right. My current names are not very Rusty...
Fixed!

PR #772, opened 2026-04-16
`slicec` Diagnostic Refactoring

InsertCreativityHere:
Right now we use the same `Diagnostic` type to report errors as we do to emit them.

However, these are really 2 different purposes, which our API reflects. `Diagnostic`s requires some post-processing before it's correct to emit them, so our API keeps them under lock-and-key, and has functions (like `into_updated`) which internally update it's fields to "convert" from one form to another.

So, instead of using 1 type for 2 purposes, with 2 different internal "states" (ready to emit **vs** not)... better to just have 2 different types, with a domain-separation approach. This PR replaces our current design by splitting `Diagnostic` in 2:
- `Diagnostic` is what the validation/parsing creates when an error or something happens.
- `PrintableDiagnostic` contains a patched & updated version of `Diagnostic` which is ready to be emitted to the user.


slicec/tests/test_helpers.rs
```diff
@@ -35,7 +35,7 @@ pub fn parse_for_ast(slice: impl Into<String>) -> Ast {
 #[must_use]
 pub fn parse_multiple_for_ast(slice: &[&str]) -> Ast {
     let compilation_state = compile_from_strings(slice, None);
-    if compilation_state.diagnostics.has_errors() {
+    if !compilation_state.diagnostics.is_empty() {
```
InsertCreativityHere, line 38: The tests should fail if there's _any_ diagnostics, not just errors.
If there were warnings in our tests, they would be silently ignored with the old logic.

slicec/tests/test_helpers.rs
```diff
@@ -115,17 +110,25 @@ pub fn check_diagnostics<const L: usize>(diagnostics: Vec<Diagnostic>, expected:
...
+            eprintln!("\texpected: \"{:?}\"", expect.span);
+            eprintln!("\t but got: \"{:?}\"", diagnostic.span);
+            failed = true;
+        }
+
+        // If a scope was provided, check that it matches.
+        if expect.scope.is_some() && expect.scope != diagnostic.scope {
+            eprintln!("diagnostic scopes didn't match:");
+            eprintln!("\texpected: \"{:?}\"", expect.scope);
+            eprintln!("\t but got: \"{:?}\"", diagnostic.scope);
```
InsertCreativityHere, line 124: We weren't checking that diagnostics were thrown the Slice scope we expected, when we really should be checking this. Thankfully there were no bugs to be caught :) But should still be checking!

slicec/examples/parse.rs
```diff
@@ -1,11 +0,0 @@
-// Copyright (c) ZeroC, Inc.
```
InsertCreativityHere, line 1: We had this outdated example laying around, from when `slicec` was library only. The example shows how to run `slicec` as a binary... But now that's actually a binary, this example is pointless.

slicec/src/diagnostics/diagnostic.rs
```diff
@@ -1,34 +1,23 @@
...
-use crate::slice_options::SliceOptions;
+use super::{DiagnosticKind, Error, Lint, Note};
+use crate::slice_file::Span;
 
 /// A diagnostic is a message that is reported to the user during compilation.
-/// It can either hold an [Error] or a [Lint].
 #[derive(Debug)]
 pub struct Diagnostic {
-    kind: DiagnosticKind,
-    level: DiagnosticLevel,
```
InsertCreativityHere, line 14: `DIagnosticLevel` is the "patched" level, which is only useful for printing the diagnostic.

i.e. a lint can become either a warning, or nothing if it's allowed.
We also plan in the future to add opt-in lints which can be upgraded to warnings.

slicec/src/parsers/mod.rs
```diff
@@ -25,7 +25,7 @@ pub fn parse_files(state: &mut CompilationState, symbols: &HashSet<String>) {
         parse_file(file, &mut state.ast, &mut diagnostics, symbols.clone());
 
         // Store any diagnostics that were emitted during parsing.
-        state.diagnostics.extend(diagnostics);
+        state.diagnostics.extend(diagnostics.into_inner());
```
InsertCreativityHere, line 28: We used to have our own `Diagnostics::extend` function, which I removed. Now we're just using the normal `Vec::extend`, but in this one spot, it requires an extra call to get the right type.

slicec/src/validators/cycle_detection.rs
```diff
@@ -147,10 +147,10 @@ impl<'a> CycleDetector<'a> {
         let cycle_notes = self.dependency_stack.iter().map(|(_, field)| Self::get_note_for(field));
 
         // Report the error.
-        Diagnostic::from_error(Error::InfiniteSizeCycle { type_id, cycle })
-            .set_span(type_being_checked.1.span())
-            .extend_notes(cycle_notes)
```
InsertCreativityHere, line 152: `extend_notes` was only used in this one spot. So I deleted `extend_notes`, and now that `DIagnostic` is all public, we can just set the `notes` field directly.

slicec/src/diagnostics/diagnostic.rs
```diff
@@ -102,36 +70,11 @@ impl Diagnostic {
...
 
-    pub fn extend_notes<I: IntoIterator<Item = Note>>(mut self, iter: I) -> Self {
-        self.notes.extend(iter);
-        self
-    }
-
     pub fn push_into(self, diagnostics: &mut Diagnostics) {
         diagnostics.0.push(self);
     }
 }
```
InsertCreativityHere, line 77: Moved `DiagnosticKind` and `DiagnosticLevel` from here (`diagnostic.rs`) to `mod.rs`. Just felt like a better place for them, and helps keep this file shorter.
Otherwise didn't change them.

slicec/src/diagnostics/diagnostic.rs
```diff
@@ -141,82 +84,29 @@ impl Diagnostics {
...
-        }
-        self.0
-    }
-
     /// Returns the diagnostics held by this without any updates or patches.
     /// This should only be called by tests that want to bypass this behavior.
     pub fn into_inner(self) -> Vec<Diagnostic> {
         self.0
     }
 }
```
InsertCreativityHere, line 99: `Deref` and `DerefMut` let you access the inner data of a wrapper type like this one.
Implementing these let's use use functions on the underlying `Vec<DIagnostic>`, instead of needing to write our own which do nothing but forward (like `is_empty` and `extend`).

slicec/src/diagnostics/diagnostic.rs
```diff
@@ -141,82 +84,29 @@ impl Diagnostics {
...
         let mut diagnostics = self.0.iter();
         diagnostics.any(|diagnostic| matches!(diagnostic.kind, DiagnosticKind::Error(_)))
     }
 
-    /// Returns true if this contains no diagnostics.
-    pub fn is_empty(&self) -> bool {
-        self.0.is_empty()
-    }
-
-    /// Returns the diagnostics this struct contains after it has patched and updated them.
```
InsertCreativityHere, line 159: Moved to `printable_diagnostic`, since this "updating" only matters for printing diagnostics to users, it doesn't matter when reporting them.

slicec/src/diagnostics/diagnostic.rs
```diff
@@ -1,34 +1,23 @@
...
-use crate::grammar::{attributes, Attributable, Entity};
-use crate::slice_file::{SliceFile, Span};
-use crate::slice_options::SliceOptions;
+use super::{DiagnosticKind, Error, Lint, Note};
+use crate::slice_file::Span;
 
 /// A diagnostic is a message that is reported to the user during compilation.
-/// It can either hold an [Error] or a [Lint].
 #[derive(Debug)]
 pub struct Diagnostic {
```
InsertCreativityHere, line 8: With the new design, `Diagnostic` is just a simple data struct.
So it's fine to make it's fields public, instead of needing `get`ters for private fields.

slicec/src/diagnostics/mod.rs
```diff
@@ -3,16 +3,37 @@
...
 
-pub use diagnostic::*;
+pub use diagnostic::{Diagnostic, Diagnostics};
 pub use errors::Error;
 pub use lints::Lint;
+pub use printable_diagnostic::*;
 
 use crate::slice_file::Span;
-use serde::Serialize;
+
```
InsertCreativityHere, line 14: `DiagnosticKind` and `DiagnosticLevel` were just moved, with no changes.

These very-small, very-central kind of types feel best suited to live in the `mod.rs` file to me.

slicec/src/validators/dictionary.rs
```diff
@@ -42,7 +42,7 @@ fn check_dictionary_key_type(type_ref: &TypeRef) -> Option<Diagnostic> {
 
                 // Convert each error into a note and add it to the struct key error.
                 for e in errors {
-                    error = error.add_note(e.message(), e.span());
+                    error = error.add_note(e.message(), e.span.as_ref());
```
InsertCreativityHere, line 45: `as_ref` is required to get the right type here.
`span()` the getter used to handle this, so think of this as: we just inlined it here.

slicec/tests/diagnostic_output_tests.rs
```diff
@@ -28,13 +28,13 @@ mod output {
 
         // Parse the Slice file.
         let state = parse(slice, Some(&options));
-        let diagnostics = state.diagnostics.into_updated(&state.ast, &state.files, &options);
+        let diagnostics = state.get_printable_diagnostics(&options);
```
InsertCreativityHere, line 31: I know it's just the tests, but you can see it's a cleaner API.
Will also be much cleaner for the vscode-slice extension when we get around to updating it!

slicec/src/lib.rs
```diff
@@ -24,6 +24,12 @@ pub fn compile_from_options(options: &SliceOptions) -> CompilationState {
     // Create an instance of `CompilationState` for holding all the compiler's state.
     let mut state = CompilationState::create();
```
InsertCreativityHere, line 26: Fixes an unrelated bug I didn't catch until this PR.
If the user requests `diable_color`, we need to be setting this option ASAP in the program, instead of just right before actually emitting them. Otherwise we may still output some color characters in the output...

slicec/src/diagnostic_emitter.rs
```diff
@@ -1,138 +1,137 @@
...
 use serde::Serializer;
 use std::io::{Result, Write};
 use std::path::Path;
 
-#[derive(Debug)]
-pub struct DiagnosticEmitter<'a, T: Write> {
+pub struct DiagnosticEmitter<'a> {
     /// Reference to the output that diagnostics should be emitted to.
-    output: &'a mut T,
+    output: &'a mut dyn Write,
```
InsertCreativityHere, line 12: Switched from using a generic `T: Write` to a dynamic `dyn Write`.
This makes the typing for `DiagnosticEmitter` easier, since it no longer has a generic type.

And in reality, it's not like we care about `DiagnosticEmitter<StdOut>` vs `DiagnosticEmitter<Vec<u8>>`...
As long as we can write to something it's fine. No need to tie the underlying write-able thing to the type of `DiagnosticEmitter`.

slicec/src/diagnostic_emitter.rs
```diff
@@ -1,138 +1,137 @@
...
     /// Can specify `json` to serialize errors as JSON or `human` to pretty-print them.
     diagnostic_format: DiagnosticFormat,
-    /// If true, diagnostic output will not be styled with colors (only used in `human` format).
-    disable_color: bool,
```
InsertCreativityHere, line 18: We disable color way earlier in the program now, instead of right before diagnostic emission.
Otherwise there were other things still being emitted in color.

slicec/src/diagnostic_emitter.rs
```diff
@@ -1,138 +1,137 @@
...
-            files,
         }
     }
 
-    pub fn emit_diagnostics(&mut self, diagnostics: Vec<Diagnostic>) -> Result<()> {
-        // Disable colors if the user requested no colors.
-        if self.disable_color {
-            console::set_colors_enabled(false);
-            console::set_colors_enabled_stderr(false);
+    pub fn get_totals(diagnostics: &[PrintableDiagnostic]) -> (usize, usize) {
```
InsertCreativityHere, line 25: This used to be a free function at the bottom of `diagnostic.rs`, now it's a static function on `DiagnosticEmitter`.
Just feels tidier to keep it here, since this was only used when emitting diagnostics.

slicec/src/diagnostic_emitter.rs
```diff
@@ -1,138 +1,137 @@
...
+                DiagnosticLevel::Error => total_errors += 1,
+                DiagnosticLevel::Warning => total_warnings += 1,
+                DiagnosticLevel::Allowed => {}
+            }
+        }
+
+        (total_warnings, total_errors)
+    }
+
+    pub fn emit_totals(total_warnings: usize, total_errors: usize) -> Result<()> {
```
InsertCreativityHere, line 39: This used to be a free function at the bottom of this file, now it's a static function on `DiagnosticEmitter`.
Just feels tidier, and saves an import in the places it's used.

slicec/tests/attribute_tests.rs
```diff
@@ -124,16 +126,17 @@ mod attributes {
                     message: "comment has a 'returns' tag, but only operations can return".to_owned(),
                 }),
             ];
```
InsertCreativityHere, line 128: With the old test setup, "allowed" lints were removed entirely.

With the new test setup, we make sure that the lints were emitted, but that they are correctly marked as `Allowed` instead of `Warning`. This feels like a safer and more true-to-task test.

slicec/tests/test_helpers.rs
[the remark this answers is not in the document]
InsertCreativityHere: This is an intentional change. In tests, we should be silently ignoring any warnings we get.
There aren't any right now, thankfully, but conceptually _all_ diagnostics should be expected in the tests.
Not just errors.

slicec/tests/attribute_tests.rs
```diff
@@ -124,16 +126,17 @@ mod attributes {
...
-                expected_indexes.contains(&(index - 1))
-            });
-            let expected: [Diagnostic; L] = all_lints.try_into().unwrap();
-
-            // Check that only the correct warnings were emitted.
-            check_diagnostics(diagnostics, expected);
+            for (index, lint) in updated_diagnostics.iter().enumerate() {
+                assert!(lint.message == all_expected_lints[index].message());
+                assert!(lint.code == all_expected_lints[index].code());
+
```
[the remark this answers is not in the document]
InsertCreativityHere: This is just a test.
If `updated_diagnostics.len() > all_expected_lints.len()` we have big problems, and it's find for the test to die.

slicec/src/diagnostics/diagnostic.rs
```diff
@@ -102,121 +70,43 @@ impl Diagnostic {
...
+/// Stores a collection of diagnostics.
 #[derive(Debug, Default)]
 pub struct Diagnostics(Vec<Diagnostic>);
 
 impl Diagnostics {
-    /// Creates a new diagnostics container that is empty.
+    /// Creates a new (empty) collection of diagnostics.
     pub fn new() -> Self {
         Self::default()
     }
```
externl, line 87: Do we really need this extra constructor, why not just use `::default`?

slicec/src/diagnostic_emitter.rs
```diff
@@ -1,138 +1,137 @@
 // Copyright (c) ZeroC, Inc.
 
-use crate::diagnostics::{Diagnostic, DiagnosticLevel};
-use crate::slice_file::{SliceFile, Span};
+use crate::diagnostics::{DiagnosticLevel, PrintableDiagnostic, Snippet};
 use crate::slice_options::{DiagnosticFormat, SliceOptions};
 use serde::ser::SerializeStruct;
 use serde::Serializer;
```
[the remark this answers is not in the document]
InsertCreativityHere: We need this on line 108 of this file. Not for the type, but for the trait methods it pulls in.
Copilot doesn't seem to understand the necessity of `use` for trait types in Rust.

slicec/src/diagnostics/mod.rs
```diff
@@ -3,16 +3,37 @@
...
+#[derive(Debug)]
+pub enum DiagnosticKind {
+    Error(Error),
+    Lint(Lint),
+}
+
+/// Diagnostic levels describe the severity of a diagnostic, and how the compiler should react to their emission.
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub enum DiagnosticLevel {
+    #[rustfmt::skip] // See https://github.com/rust-lang/rustfmt/issues/5801
```
externl, line 24: This issue is closed, presumably the fix should be in the nighty.

slicec/src/diagnostics/printable_diagnostic.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub code: String,
+    pub snippet: Option<Snippet>,
+    pub notes: Vec<PrintableNote>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct PrintableNote {
+    pub message: String,
+    #[serde(rename = "span")]
+    pub snippet: Option<Snippet>,
```
externl: Why do we encode as "span" in JSON?

slicec/src/diagnostics/printable_diagnostic.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct Snippet {
+    pub span: Span,
+    pub text: String,
+}
+// TODO should we send the text snippet in JSON diagnostic format? Or just the range?
```
externl: I would just include it for simplicity. Right now all our Snippets are really just spans in JSON.

slicec/src/diagnostic_emitter.rs
```diff
@@ -1,138 +1,137 @@
...
 
             let mut serializer = serde_json::Serializer::new(&mut *self.output);
             let mut state = serializer.serialize_struct("Diagnostic", 5)?;
-            state.serialize_field("message", &diagnostic.message())?;
+            state.serialize_field("message", &diagnostic.message)?;
             state.serialize_field("severity", severity)?;
-            state.serialize_field("span", &diagnostic.span())?;
-            state.serialize_field("notes", diagnostic.notes())?;
-            state.serialize_field("error_code", diagnostic.code())?;
+            state.serialize_field("span", &diagnostic.snippet)?;
```
externl: I think we should either encode just the span like we did before (but as diagnostic.snippet.span`, or encode the entire snippet.

slicec/src/diagnostics/printable_diagnostic.rs
InsertCreativityHere: This makes it backwards compatible with the JSON decoding the tooling does.

But, it depended on what you answered to the TODO :) I'll update to send the full snippet.

slicec/src/diagnostic_emitter.rs
InsertCreativityHere: This PR currently does option 1.
If you look at the implementation where `snippet` is encoded, all it actually encodes is the span.

Personally, I'd lean towards option 2 (encode the full snippet (span + text),
but wasn't sure what other opinions would be.

slicec/src/diagnostics/mod.rs
InsertCreativityHere: This was just copy-pasted. Didn't even think to check, good idea.

slicec/src/diagnostics/diagnostic.rs
InsertCreativityHere: > Do we really need this extra constructor

Honestly, no.

> why not just use ::default?

It just feels more semantically correct to create a 'new' `Diagnostic`, instead of a 'default' `Diagnostic`.
That was my feeling at least. If you don't like it though, I can remove or simplify! : ^)

slicec/src/diagnostics/mod.rs
InsertCreativityHere: Doesn't seem like it's actually fixed? Will look into, separate from this PR ;-;

InsertCreativityHere:
I got rid of my little hacks to make the JSON diagnostics keep their current encoding.
Now, we send along the text snippet in the JSON, whereas previously, the text as only printed to the console.

I'll have to update the Task which consumes these diagnostics to handle this extra field.

slicec/src/diagnostics/printable_diagnostic.rs
```diff
@@ -0,0 +1,111 @@
...
+        snippet: get_snippet(&n.span, &compilation_state.files),
+    });
+
+    PrintableDiagnostic {
+        message: diagnostic.message(),
+        level: get_diagnostic_level_for(diagnostic, options, compilation_state),
+        code: diagnostic.code().to_owned(),
+        snippet: get_snippet(&diagnostic.span, &compilation_state.files),
+        notes: notes.collect(),
+    }
```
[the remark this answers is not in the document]
InsertCreativityHere: This is fine. Extracting the snippets up-front instead of lazily on demand makes the code alot simpler, and we expect the vast majority of lints and errors will _not_ be suppressed. So in reality, the performance savings here are minimal.

slicec/src/diagnostics/printable_diagnostic.rs
```diff
@@ -0,0 +1,111 @@
...
+pub struct PrintableDiagnostic {
+    pub message: String,
+    pub level: DiagnosticLevel,
+    pub code: String,
+    pub snippet: Option<Snippet>,
+    pub notes: Vec<PrintableNote>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct PrintableNote {
```
externl, line 22: What about something like `AnnotatedDiagnostic` and `AnnotatedNote`?

slicec/src/diagnostics/printable_diagnostic.rs
```diff
@@ -0,0 +1,111 @@
...
+        message: diagnostic.message(),
+        level: get_diagnostic_level_for(diagnostic, options, compilation_state),
+        code: diagnostic.code().to_owned(),
+        snippet: get_snippet(&diagnostic.span, &compilation_state.files),
+        notes: notes.collect(),
+    }
+}
+
+/// Returns the [`DiagnosticLevel`] that the provided [`Diagnostic`] should be emitted with.
+fn get_diagnostic_level_for(
```
externl, line 54: what about `effective_diagnostic_level`?

externl (approved):
Looks good! See comments.

PR #775, opened 2026-05-07
Finished Adding all the Diagnostics to the Slice-Compiler Definitions

InsertCreativityHere:
This is the penultimate PR in the complete refactoring of `slicec`'s diagnostic system, so it plays well with the code-generators.
It reworks the Slice-Compiler Definitions for Diagnostic, most notably, by switching to a `DiagnosticKind` enum, instead of allowing `Diagnostic` to hold any arbitrary message.

This:
1) Makes error messages more standard, since they're always created by `slicec` now, instead of each plugin needing to create and format it's own strings.
2) Allows plugin issued diagnostics to work with the `allow/deny` lint configuration system (not possible when all `slicec` saw was a string)
3) Allows notes to be directly attached to a diagnostic, instead of `slicec` needing to try and infer whether a diagnostic is linked to the one that came before it now.

----

This just defines the new structure, my next (and final) PR will actually hook up `slicec` to receive and convert plug-in diagnostics into it's own native diagnostic types.

### DIFF
--- a/slice/Compiler/CodeGenerator.slice
+++ b/slice/Compiler/CodeGenerator.slice
@@ -33,15 +33,3 @@ enum Symbol {
     DictionaryType(v: DictionaryType)
     TypeAlias(v: TypeAlias)
 }
-
-struct Diagnostic {
-    level: DiagnosticLevel
-    message: string
-    source: string?
-}
-
-unchecked enum DiagnosticLevel : uint8 {
-    Info
-    Warning
-    Error
-}
--- a/slice/Compiler/Diagnostics.slice
+++ b/slice/Compiler/Diagnostics.slice
@@ -1,0 +1,74 @@
+// Copyright (c) ZeroC, Inc.
+
+[cs::identifier("ZeroC.Slice.Symbols.Compiler")]
+module Compiler
+
+/// A message that is reported to provide information and context for errors, warnings, or other issues.
+struct Diagnostic {
+    /// The exact kind of diagnostic.
+    kind: DiagnosticKind
+
+    /// The fully-scoped Slice identifier of the element that this diagnostic is in reference to (if there is one).
+    source: string?
+
+    /// Any additional information that should be reported alongside this diagnostic's main message.
+    notes: Sequence<DiagnosticNote>
+}
+
+/// Stores additional information to report alongside a {@link Diagnostic}.
+struct DiagnosticNote {
+    /// The message to emit alongside the main diagnostic which this note is attached to.
+    message: string
+
+    /// The fully-scoped Slice identifier of the element that this note is in reference to (if there is one).
+    source: string?
+}
+
+unchecked enum DiagnosticKind {
+    /// A general purpose diagnostic which should be emitted as an error.
+    Error(message: string)
+
+    /// A general purpose diagnostic which should be emitted as a warning.
+    Warning(message: string)
+
+    /// A general purpose diagnostic which is purely informational.
+    Info(message: string)
+
+    //
+    // Attribute Diagnostics
+    //
+
+    /// An attribute was applied to a Slice element for which it's invalid.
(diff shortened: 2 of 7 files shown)


slice/Compiler/CodeGenerator.slice
```diff
@@ -33,15 +33,3 @@ enum Symbol {
     DictionaryType(v: DictionaryType)
     TypeAlias(v: TypeAlias)
 }
-
-struct Diagnostic {
```
InsertCreativityHere, line 37: Moved all the diagnostic stuff into it's own Slice file.
It always felt a little weird having them attached to `CodeGenerator` to me,

slicec/src/main.rs
```diff
@@ -114,7 +114,7 @@ fn handle_generator_response(response_payload: Vec<u8>, output_dir: &Option<Stri
     // TODO: Convert the diagnostics we decode from the generator, into diagnostics that slicec can handle.
     //       To do this requires re-working the diagnostic API fairly substantially.
     for generator_diagnostic in generator_diagnostics {
-        println!("{}", generator_diagnostic.message);
```
InsertCreativityHere, line 117: The spot that will finally be fixed in my next PR.
For now, just adapted the placeholder code.

slicec/src/definition_types.rs
```diff
@@ -10,7 +10,7 @@ use slice_codec::decode_from::DecodeFrom;
 use slice_codec::decoder::Decoder;
 use slice_codec::encode_into::EncodeInto;
```
InsertCreativityHere, line 11: This file is just the hand-written encoding/decoding code for the Slice compiler definitions.
Eventually will be replaced with generated code.

slicec/src/diagnostics/lints.rs
```diff
@@ -22,24 +26,31 @@ pub enum Lint {
     },
 
     /// A syntactical mistake in a doc-comment.
```
InsertCreativityHere, line 28: Who knows why, but for some reason adding the `Other` variant makes `rustfmt` prefer these other 3 variants be written on multiple lines.

slicec/src/diagnostics/diagnostic.rs
```diff
@@ -44,10 +44,10 @@ impl Diagnostic {
         Self::new(DiagnosticKind::Lint(lint))
     }
 
-    /// Creates a new informational `Diagnostic` from the provided string.
+    /// Creates a new informational `Diagnostic` from the provided message.
     /// The newly created `Diagnostic` has no `span`, `scope`, or `notes` set.
-    pub fn info(info: impl Into<String>) -> Self {
```
InsertCreativityHere, line 49: Renamed to be consistent with the other `from_error` and `from_lint` constructors.

slice/Compiler/Diagnostics.slice
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) ZeroC, Inc.
+
```
InsertCreativityHere, line 2: This file is the real bulk of this PR.
It is all new, and replaces the older diagnostic format you can see in `CodeGenerator.slice`.

slicec/src/definition_types.rs
```diff
@@ -322,47 +322,125 @@ impl DecodeFrom for GeneratedFile {
...
+                min_expected: decoder.decode()?,
+                max_expected: decoder.decode()?,
+                actual_count: decoder.decode()?,
+            },
+
+            n => Self::Unknown { discriminant: n },
+        };
+
+        decoder.skip_tagged_fields()?;
+        Ok(variant)
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed. This was indeed a bug in my decoding logic.

pepone (approved):

externl (approved):
Looks good!

PR #776, opened 2026-05-11
Properly Handle Diagnostics From Plugins

### LINKED ISSUE #771: Ensure `slicec` Emits JSON Diagnostics even upon Generator Failure
Right now these kinds of error messages are just printed to `stderr`, even when JSON output has been requested.
This causes tools which expect JSON to fail in unreadable ways. For example see: https://github.com/icerpc/icerpc-csharp/issues/4504

InsertCreativityHere:
This PR is the Rust counterpart to https://github.com/icerpc/icerpc-csharp/pull/4615.
It fixes #771.

It adds the logic to convert from a slice-generated `Compiler::Diagnostic` to a native `slicec` `Diagnostic`,
and updates the `main` function to properly decode, convert, process, and emit diagnostics reported by code-generators.

----

Running `slicec` by hand on the CLI shows it's working good. But of course the real test will be once this is merged, and can be used by `icerpc-csharp` directly. This at least shows that the decoding/conversion is all correct.
```
error [E023]: invalid attribute 'cs::encodedReturn'
 --> .\icerpc-csharp\Test.slice:2:3
  |
2 | [[cs::encodedReturn]]
  |   -----------------
  |
error [E024]: unknown attribute 'cs::whatever'
 --> .\icerpc-csharp\Test.slice:6:2
  |
6 | [cs::whatever]
  |  ------------
  |
error [E028]: attribute 'cs::identifier' takes exactly 1 argument, but 2 arguments were provided
 --> .\icerpc-csharp\Test.slice:8:6
  |
8 |     [cs::identifier("Foo", "Bar")]
  |      ----------------------------
  |
error [E025]: missing required attribute 'cs::type(TYPE_STRING)'
 --> .\icerpc-csharp\Test.slice:12:1
   |
12 | custom MyType
   | -------------
   |
Failed: Compilation failed with 4 error(s)
```

### DIFF
--- a/slice/Compiler/Diagnostics.slice
+++ b/slice/Compiler/Diagnostics.slice
@@ -9,7 +9,7 @@ module Compiler
 /// - A reference to a Slice entity, written as its fully-scoped identifier; ex: 'MyModule::MyInterface::myOp'.
 /// - A reference to a Slice file, written as that file's relative path prefixed with '#'; ex: '#dir1/dir2/file.slice'.
 ///
-/// To reference an attribute on a Slice file or entity, you can place '$attributes' after the source, followed by the
+/// To reference an attribute on a Slice entity or file, you can use '$attributes' after the source, followed by the
 /// index of the attribute to reference, and optionally, the index of an argument to that attribute. For example:
 /// 'MyMod::MyEnum::$attributes::0' references the first attribute of 'MyEnum', and '#MyFile.slice::$attributes::0::1'
 /// references the second argument of the first attribute on 'MyFile.slice'.
--- a/slice/Compiler/Diagnostics.slice
+++ b/slice/Compiler/Diagnostics.slice
@@ -13,7 +13,7 @@ module Compiler
 /// index of the attribute to reference, and optionally, the index of an argument to that attribute. For example:
 /// 'MyMod::MyEnum::$attributes::0' references the first attribute of 'MyEnum', and '#MyFile.slice::$attributes::0::1'
 /// references the second argument of the first attribute on 'MyFile.slice'.
-typealias Source = string
+typealias DiagnosticSource = string
 
 /// A message that is reported to provide information and context for errors, warnings, or other issues.
 struct Diagnostic {
@@ -21,7 +21,7 @@ struct Diagnostic {
     kind: DiagnosticKind
 
     /// The fully-scoped Slice identifier of the element that this diagnostic is in reference to (if there is one).
-    source: Source?
+    source: DiagnosticSource?
 
     /// Any additional information that should be reported alongside this diagnostic's main message.
     notes: Sequence<DiagnosticNote>
@@ -33,7 +33,7 @@ struct DiagnosticNote {
     message: string
 
     /// The fully-scoped Slice identifier of the element that this note is in reference to (if there is one).
-    source: Source?
+    source: DiagnosticSource?
 }
 
 unchecked enum DiagnosticKind {
--- a/slice/Compiler/Diagnostics.slice
+++ b/slice/Compiler/Diagnostics.slice
@@ -6,11 +6,11 @@ module Compiler
 /// Represents the source of a {@link Diagnostic}, which can be used to provide context and snippets during emission.
 ///
 /// The source of a diagnostic can be either:
-/// - A reference to a Slice entity, written as it's fully-scoped identifier; ex: 'MyModule::MyInterface::myOp'.
+/// - A reference to a Slice entity, written as its fully-scoped identifier; ex: 'MyModule::MyInterface::myOp'.
 /// - A reference to a Slice file, written as that file's relative path prefixed with '#'; ex: '#dir1/dir2/file.slice'.
 ///
-/// To reference an attribute on a Slice file or entity, you can place '$attributes' after the source, following by the
-/// the index of the attribute to reference, and optionally, the index of an argument to that attribute. For example:
+/// To reference an attribute on a Slice file or entity, you can place '$attributes' after the source, followed by the
+/// index of the attribute to reference, and optionally, the index of an argument to that attribute. For example:
 /// 'MyMod::MyEnum::$attributes::0' references the first attribute of 'MyEnum', and '#MyFile.slice::$attributes::0::1'
 /// references the second argument of the first attribute on 'MyFile.slice'.
 typealias Source = string
--- a/slice/Compiler/Diagnostics.slice
+++ b/slice/Compiler/Diagnostics.slice
@@ -3,13 +3,16 @@
 [cs::identifier("ZeroC.Slice.Symbols.Compiler")]
 module Compiler
 
+/// TODO
+typealias Source = string
+
 /// A message that is reported to provide information and context for errors, warnings, or other issues.
 struct Diagnostic {
     /// The exact kind of diagnostic.
     kind: DiagnosticKind
 
     /// The fully-scoped Slice identifier of the element that this diagnostic is in reference to (if there is one).
-    source: string?
+    source: Source?
 
     /// Any additional information that should be reported alongside this diagnostic's main message.
     notes: Sequence<DiagnosticNote>
@@ -21,7 +24,7 @@ struct DiagnosticNote {
     message: string
 
     /// The fully-scoped Slice identifier of the element that this note is in reference to (if there is one).
-    source: string?
+    source: Source?
 }
 
 unchecked enum DiagnosticKind {
--- a/slice/Compiler/Diagnostics.slice
+++ b/slice/Compiler/Diagnostics.slice
@@ -3,7 +3,16 @@
 [cs::identifier("ZeroC.Slice.Symbols.Compiler")]
 module Compiler
 
-/// TODO
+/// Represents the source of a {@link Diagnostic}, which can be used to provide context and snippets during emission.
+///
+/// The source of a diagnostic can be either:
+/// - A reference to a Slice entity, written as it's fully-scoped identifier; ex: 'MyModule::MyInterface::myOp'.
+/// - A reference to a Slice file, written as that file's relative path prefixed with '#'; ex: '#dir1/dir2/file.slice'.
+///
+/// To reference an attribute on a Slice file or entity, you can place '$attributes' after the source, following by the
+/// the index of the attribute to reference, and optionally, the index of an argument to that attribute. For example:
+/// 'MyMod::MyEnum::$attributes::0' references the first attribute of 'MyEnum', and '#MyFile.slice::$attributes::0::1'
+/// references the second argument of the first attribute on 'MyFile.slice'.
 typealias Source = string
 
 /// A message that is reported to provide information and context for errors, warnings, or other issues.
--- a/slicec/src/ast/mod.rs
+++ b/slicec/src/ast/mod.rs
@@ -366,3 +366,13 @@ pub enum LookupError {
         is_concrete: bool,
     },
 }
+
+impl From<LookupError> for crate::diagnostics::Error {
+    fn from(error: LookupError) -> Self {
+        match error {
+            LookupError::DoesNotExist { identifier } => Self::DoesNotExist { identifier },
+            LookupError::TypeMismatch { expected, actual, is_concrete }
+                => Self::TypeMismatch { expected, actual, is_concrete },
+        }
+    }
+}
--- a/slicec/src/ast/mod.rs
+++ b/slicec/src/ast/mod.rs
@@ -371,8 +371,15 @@ impl From<LookupError> for crate::diagnostics::Error {
     fn from(error: LookupError) -> Self {
         match error {
             LookupError::DoesNotExist { identifier } => Self::DoesNotExist { identifier },
-            LookupError::TypeMismatch { expected, actual, is_concrete }
-                => Self::TypeMismatch { expected, actual, is_concrete },
+            LookupError::TypeMismatch {
+                expected,
+                actual,
+                is_concrete,
+            } => Self::TypeMismatch {
+                expected,
+                actual,
+                is_concrete,
+            },
         }
     }
 }
--- a/slicec/src/ast/node.rs
+++ b/slicec/src/ast/node.rs
@@ -111,6 +111,34 @@ impl<'a> TryFrom<&'a Node> for WeakPtr<dyn Type> {
     }
 }
 
+impl<'a> TryFrom<&'a Node> for &'a dyn NamedSymbol {
+    type Error = LookupError;
+
+    /// Attempts to unwrap a node to a dynamically typed reference of a Slice [NamedSymbol].
+    ///
+    /// If the Slice element held by the node implements [NamedSymbol], this succeeds and returns a typed reference,
+    /// otherwise this fails and returns an error message.
+    fn try_from(node: &'a Node) -> Result<&'a dyn NamedSymbol, Self::Error> {
+        match node {
+            Node::Module(module_ptr) => Ok(module_ptr.borrow()),
+            Node::Struct(struct_ptr) => Ok(struct_ptr.borrow()),
+            Node::Field(field_ptr) => Ok(field_ptr.borrow()),
+            Node::Interface(interface_ptr) => Ok(interface_ptr.borrow()),
+            Node::Operation(operation_ptr) => Ok(operation_ptr.borrow()),
+            Node::Parameter(parameter_ptr) => Ok(parameter_ptr.borrow()),
+            Node::Enum(enum_ptr) => Ok(enum_ptr.borrow()),
+            Node::Enumerator(enumerator_ptr) => Ok(enumerator_ptr.borrow()),
+            Node::CustomType(custom_type_ptr) => Ok(custom_type_ptr.borrow()),
+            Node::TypeAlias(type_alias_ptr) => Ok(type_alias_ptr.borrow()),
+            _ => Err(LookupError::TypeMismatch {
+                expected: "named symbol".to_owned(),
+                actual: ccase!(lower, node.to_string()),
+                is_concrete: false,
+            }),
+        }
+    }
+}
+
 impl<'a> TryFrom<&'a Node> for &'a dyn Entity {
     type Error = LookupError;
 
--- a/slicec/src/compilation_state.rs
+++ b/slicec/src/compilation_state.rs
@@ -34,7 +34,7 @@ impl CompilationState {
     ///
     /// # Safety
     ///
-    /// The caller of this function must ensure that no (`WeakPtr`s)[crate::utils::ptr_util::WeakPtr] exist that point
+    /// The caller of this function must ensure that no [`WeakPtr`](crate::utils::ptr_util::WeakPtr)s exist that point
     /// to the contents of this `CompilationState`. Even if they're not being actively used, their existence causes UB.
     pub unsafe fn apply_unsafe(&mut self, function: unsafe fn(&mut Self)) {
         if !self.diagnostics.has_errors() {
--- a/slicec/src/compilation_state.rs
+++ b/slicec/src/compilation_state.rs
@@ -45,7 +45,7 @@ impl CompilationState {
     pub fn get_annotated_diagnostics(&self, options: &SliceOptions) -> Vec<AnnotatedDiagnostic> {
         self.diagnostics
             .iter()
-            .map(|diagnostic| crate::diagnostics::convert_diagnostic(diagnostic, options, self))
+            .map(|diagnostic| crate::diagnostics::convert_diagnostic(diagnostic, options, &self.ast, &self.files))
             .collect()
     }
 }
--- a/slicec/src/definition_types.rs
+++ b/slicec/src/definition_types.rs
@@ -405,6 +405,8 @@ pub enum DiagnosticKind {
 impl DecodeFrom for DiagnosticKind {
     fn decode_from(decoder: &mut Decoder<impl InputSource>) -> Result<Self> {
         let value: usize = decoder.decode_varint()?;
+        let payload_size = decoder.decode_size()?;
+
         let variant = match value {
             0 => Self::Info {
                 message: decoder.decode()?,
@@ -441,7 +443,6 @@ impl DecodeFrom for DiagnosticKind {
             n => {
                 // Read the unknown variant's fields payload.
                 // We don't know what they are, but we at least know the length of the payload.
-                let payload_size = decoder.decode_size()?;
                 let mut fields_payload = vec![0; payload_size];
                 decoder.read_bytes_into_exact(&mut fields_payload)?;
 
--- a/slicec/src/diagnostics/annotated_diagnostic.rs
+++ b/slicec/src/diagnostics/annotated_diagnostic.rs
@@ -1,6 +1,6 @@
 // Copyright (c) ZeroC, Inc.
 
-use crate::compilation_state::CompilationState;
+use crate::ast::Ast;
 use crate::diagnostics::{Diagnostic, DiagnosticKind, DiagnosticLevel, Lint};
 use crate::grammar::{attributes, Attributable, Entity};
 use crate::slice_file::{SliceFile, Span};
@@ -34,18 +34,19 @@ pub struct Snippet {
 pub fn convert_diagnostic(
     diagnostic: &Diagnostic,
     options: &SliceOptions,
-    compilation_state: &CompilationState,
+    ast: &Ast,
+    files: &[SliceFile],
 ) -> AnnotatedDiagnostic {
     let notes = diagnostic.notes.iter().map(|n| AnnotatedNote {
         message: n.message.clone(),
-        snippet: get_snippet(&n.span, &compilation_state.files),
+        snippet: get_snippet(&n.span, files),
     });
 
     AnnotatedDiagnostic {
         message: diagnostic.message(),
-        level: get_diagnostic_level_for(diagnostic, options, compilation_state),
+        level: get_diagnostic_level_for(diagnostic, options, ast, files),
         code: diagnostic.code().to_owned(),
-        snippet: get_snippet(&diagnostic.span, &compilation_state.files),
+        snippet: get_snippet(&diagnostic.span, files),
         notes: notes.collect(),
     }
 }
@@ -54,7 +55,8 @@ pub fn convert_diagnostic(
 fn get_diagnostic_level_for(
     diagnostic: &Diagnostic,
     options: &SliceOptions,
-    compilation_state: &CompilationState,
+    ast: &Ast,
+    files: &[SliceFile],
 ) -> DiagnosticLevel {
     // Only lints can have their diagnostic levels changed (through attributes or command-line options).
     // For other kinds of diagnostics, we can immediately return their levels.
@@ -83,7 +85,7 @@ fn get_diagnostic_level_for(
 
     // If the diagnostic has a span, check if it's affected by an `allow` attribute on its file.
     if let Some(span) = &diagnostic.span {
-        let file = compilation_state.files.iter().find(|f| f.relative_path == span.file);
+        let file = files.iter().find(|f| f.relative_path == span.file);
         if is_lint_allowed_by_attributes(file.unwrap(), lint) {
             return DiagnosticLevel::Allowed;
         }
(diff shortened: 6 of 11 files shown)


slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,241 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+                return (None, None);
+            }
+        };
+
+        // Determine which 'span' to return based off the optional extension.
+        // The 'scope' is always `None`, since files do not logically have a Slice scope like symbols do.
+        let scope = None;
+        let span = match extension {
+            // If the extension starts with 'attributes::', then this diagnostic is referencing the symbol's attributes.
+            Some("attributes::") => get_attribute_span(slice_file, extension.unwrap(), output),
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed.

slicec/src/main.rs
```diff
@@ -75,6 +79,11 @@ fn spawn_plugin_process(plugin: &Plugin, slice_payload: &[u8]) -> std::io::Resul
     Ok(subprocess)
 }
 
+/// Runs the provided subprocess to completion, handling any failures that may occur during its execution.
+///
+/// If the subprocess completes successfully, this returns `Ok` with it's response payload.
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed.

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,241 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+
+        // Determine which 'span' to return based off the optional extension.
+        // The 'scope' is always the fully-scoped identifier of the symbol, since meta-elements don't have Slice scopes.
+        let scope = Some(named_symbol.parser_scoped_identifier());
+        let span = match extension {
+            // If there was no extension, this diagnostic is referencing the named symbol itself.
+            None => Some(named_symbol.span().clone()),
+
+            // If the extension starts with 'attributes::', then this diagnostic is referencing the symbol's attributes.
+            Some("attributes::") => get_attribute_span(named_symbol, extension.unwrap(), output),
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed like the other one.

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,241 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+    // Split the extension string by '::' and ensure it starts with 'attributes' and has at least 2 parts.
+    let indices = extension.split("::").collect::<Vec<_>>();
+    assert!(indices.len() > 1 && indices[0] == "attributes");
+
+    // Make sure we either 1 or 2 indices after 'attributes'.
+    let (attribute_index_str, argument_index_str) = match &indices[1..] {
+        [i] => (i.parse::<usize>(), None),
+        [i, j] => (i.parse::<usize>(), Some(j.parse::<usize>())),
+
+        [] => unreachable!("'get_attribute_span' had 0 indices despite asserting that there was at least 1!"),
```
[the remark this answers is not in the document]
InsertCreativityHere: This assert is correct, and this function only gets called when the assert would hold.
This is on us to ensure, and we aren't accepting an unstructured string in this function.

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,241 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+fn get_attribute_span<T: Attributable + ?Sized>(symbol: &T, extension: &str, output: &mut Diagnostics) -> Option<Span> {
+    // Split the extension string by '::' and ensure it starts with 'attributes' and has at least 2 parts.
+    let indices = extension.split("::").collect::<Vec<_>>();
+    assert!(indices.len() > 1 && indices[0] == "attributes");
+
+    // Make sure we either 1 or 2 indices after 'attributes'.
+    let (attribute_index_str, argument_index_str) = match &indices[1..] {
+        [i] => (i.parse::<usize>(), None),
+        [i, j] => (i.parse::<usize>(), Some(j.parse::<usize>())),
+
```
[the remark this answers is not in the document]
InsertCreativityHere: Will fix in a follow-up PR after refactoring attributes,

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,241 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+    // This optional extension always begins with a '$' and can be used to refer to meta-elements attached to the symbol
+    // like attributes or doc-comments. For example: `"MyModule::MyClass::$attributes::1"` for attribute 1 on "MyClass".
+    let (entity_id, extension) = if let Some((entity_id, extension)) = source.split_once("::$") {
+        (entity_id, Some(extension))
+    } else {
+        (source, None)
+    };
+
+    // If the source starts with a '#', then it is referencing a file. Otherwise it is referencing a named symbol.
+    if let Some(diagnostic_file_path) = source.strip_prefix('#') {
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed.

slicec/src/ast/mod.rs
```diff
@@ -366,3 +366,20 @@ pub enum LookupError {
         is_concrete: bool,
     },
 }
+
+impl From<LookupError> for crate::diagnostics::Error {
```
InsertCreativityHere, line 370: Just a helper to make it easier to convert from `LookupError`s (failed to lookup something in the AST) to our main `Error` type. We do this conversion in a couple places, so better to have a single helper.

slice/Compiler/Diagnostics.slice
```diff
@@ -3,13 +3,25 @@
 [cs::identifier("ZeroC.Slice.Symbols.Compiler")]
 module Compiler
 
+/// Represents the source of a {@link Diagnostic}, which can be used to provide context and snippets during emission.
```
InsertCreativityHere, line 6: Added a new `Source` type to communicate "this can't be any random string", and so we have a place to document the structure of it.

slicec/src/diagnostics/annotated_diagnostic.rs
```diff
@@ -34,18 +34,19 @@ pub struct Snippet {
 pub fn convert_diagnostic(
     diagnostic: &Diagnostic,
     options: &SliceOptions,
-    compilation_state: &CompilationState,
```
InsertCreativityHere, line 37: Instead of passing the entire `CompilationState`, now we only pass the specific fields we need.
This is a little cleaner I think.

slicec/src/diagnostics/errors.rs
```diff
@@ -387,9 +387,8 @@ impl Error {
                     if expected == 0 {
                         format!("attribute '{directive}' does not take any arguments, but {args_provided}")
                     } else {
```
InsertCreativityHere, line 389: This caused us to emit messages like:
`attribute 'cs::type' takes exactly 1 argument, but only 0 arguments were provided`
"but only 0" is a weird thing to say. Better to just keep it simpler, and never emit the "only".

slicec/src/definition_types.rs
```diff
@@ -405,6 +405,8 @@ pub enum DiagnosticKind {
 impl DecodeFrom for DiagnosticKind {
     fn decode_from(decoder: &mut Decoder<impl InputSource>) -> Result<Self> {
         let value: usize = decoder.decode_varint()?;
```
InsertCreativityHere, line 407: This was a straight-up bug in my implementation.
For unchecked `enum`s, we _always_ encode the payload size up-front, not just for "Unknown" values.

Now we always decode this size, after decoding the discriminant. Even if it's only used in the case of an `Unknown` variant (to skip it).

slicec/src/compilation_state.rs
```diff
@@ -45,7 +45,7 @@ impl CompilationState {
     pub fn get_annotated_diagnostics(&self, options: &SliceOptions) -> Vec<AnnotatedDiagnostic> {
         self.diagnostics
             .iter()
-            .map(|diagnostic| crate::diagnostics::convert_diagnostic(diagnostic, options, self))
+            .map(|diagnostic| crate::diagnostics::convert_diagnostic(diagnostic, options, &self.ast, &self.files))
```
InsertCreativityHere: Where we only pass the fields we need instead of the entire `CompilationState`.

slicec/src/compilation_state.rs
```diff
@@ -34,7 +34,7 @@ impl CompilationState {
     ///
     /// # Safety
     ///
-    /// The caller of this function must ensure that no (`WeakPtr`s)[crate::utils::ptr_util::WeakPtr] exist that point
+    /// The caller of this function must ensure that no [`WeakPtr`](crate::utils::ptr_util::WeakPtr)s exist that point
```
InsertCreativityHere, line 37: I got the Markdown `[]` and `()` backwards here and just happened to notice...
I can never remember the right order...

slicec/src/main.rs
```diff
@@ -105,39 +115,33 @@ fn collect_plugin_output(subprocess: Child) -> std::io::Result<Vec<u8>> {
...
-    // TODO: Convert the diagnostics we decode from the generator, into diagnostics that slicec can handle.
-    //       To do this requires re-working the diagnostic API fairly substantially.
+    // Take all of the decoded diagnostics from the generator, and convert them into diagnostics that slicec can handle.
+    let mut converted_diagnostics = Diagnostics::new();
     for generator_diagnostic in generator_diagnostics {
-        println!("{generator_diagnostic:?}");
-    }
-    let mut diagnostics = Diagnostics::new();
-
-    // If no errors were reported, attempt to generate the files in the response.
```
InsertCreativityHere, line 121: This function was renamed from `handle_generator_response` to just `decode_generator_response`.
It's responsibilities no longer include "actually write the generated code to disk".
This was moved further down this file.

slicec/src/main.rs
```diff
@@ -180,14 +197,18 @@ fn main() -> ExitCode {
...
 
-    // Perform the compilation.
+    // Compile the provided Slice files.
     let mut compilation_state = slicec::compile_from_options(&slice_options);
-    let diagnostics = &mut compilation_state.diagnostics;
+    let CompilationState {
+        ref ast,
+        ref mut diagnostics,
+        ref files,
+    } = compilation_state;
```
InsertCreativityHere, line 201: Destructuring assignment, now that we need the `ast` and `files` in addition to the `diagnostics`.

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,241 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+
+    // If a 'source' was provided, convert it to a corresponding 'span' and 'scope'.
+    let (span, scope) = convert_source(diagnostic.source.as_deref(), ast, files, output);
+
+    // Construct and push the converted 'slicec' `Diagnostic`.
+    let converted_diagnostic = slicec::diagnostics::Diagnostic { kind, span, scope, notes };
+    output.push(converted_diagnostic);
+}
+
+fn convert_source(source: Option<&str>, ast: &Ast, files: &[GrammarSliceFile], output: &mut Diagnostics) -> (Option<Span>, Option<String>) {
```
InsertCreativityHere, line 120: All this functionality to parse, validate, and resolve the `source` string feels bloated to me.
But everything else I've tried produces longer code. If you have any better ways to structure this, I am all ears.

slicec/src/ast/node.rs
```diff
@@ -111,6 +111,34 @@ impl<'a> TryFrom<&'a Node> for WeakPtr<dyn Type> {
     }
 }
 
+impl<'a> TryFrom<&'a Node> for &'a dyn NamedSymbol {
```
InsertCreativityHere, line 114: This allows us to write functions for, and search the AST for (entities + Module), since modules are a little different and special.

We used to have this function a while ago, but then it was unused, so we removed it, but we once again need it to look up diagnostic sources in the AST (since modules can be referenced by sources).

slicec/src/grammar/traits.rs
```diff
@@ -22,7 +22,7 @@ pub trait ScopedSymbol: Symbol {
     fn get_raw_scope(&self) -> &Scope;
 }
 
-pub trait NamedSymbol: Symbol {
+pub trait NamedSymbol: Symbol + Attributable {
```
InsertCreativityHere, line 25: It has always been the case that all `NamedSymbol`s are `Attributable`.
But we never added this because it was simply not necessary before.

Perhaps some future refactoring around modules, or how module attributes work can let us once again remove this coupling.

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,242 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+            }
+
+            None => {
+                let message = format!("the diagnostic source '{source}' is missing a required source extension");
+                let error = SlicecDiagnostic::from_error(SlicecError::Other { message })
+                    .add_note(CRITICAL_FLAW_STRING, None)
+                    .add_note("file-sources must have an extension", None);
+                output.push(error);
+                None // There is no meaningful span to return.
+            }
```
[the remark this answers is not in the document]
InsertCreativityHere: Already fixed. We just point to the start of the file in this case now. Better than returning `None`.

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,242 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+}
+
+/// Parses a diagnostic 'source' string, and returns the corresponding 'span' and 'scope' for the referenced element.
+fn convert_source(source: Option<&str>, ast: &Ast, files: &[GrammarSliceFile], output: &mut Diagnostics) -> (Option<Span>, Option<String>) {
+    // If no source was provided, we can immediately return `(None, None)` since there's nothing to convert.
+    let Some(source) = source else {
+        return (None, None);
+    };
+
+    // Otherwise, split the provided source into 'the scoped id of an symbol' and an 'optional extension'.
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed.

slicec/src/definition_types.rs
```diff
@@ -405,6 +405,8 @@ pub enum DiagnosticKind {
...
+        let payload_size = decoder.decode_size()?;
+
```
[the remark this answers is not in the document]
InsertCreativityHere: It's just wrong here. We do use `payload_size` when we encounter an unknown discriminant and need to skip it.

slice/Compiler/Diagnostics.slice
```diff
@@ -3,13 +3,25 @@
...
+///
+/// The source of a diagnostic can be either:
+/// - A reference to a Slice entity, written as its fully-scoped identifier; ex: 'MyModule::MyInterface::myOp'.
+/// - A reference to a Slice file, written as that file's relative path prefixed with '#'; ex: '#dir1/dir2/file.slice'.
```
externl, line 10: Maybe we can use file URIs here. For example: `file://dir1/dir2/file.slice`.

EDIT: see other comment.

slice/Compiler/Diagnostics.slice
```diff
@@ -3,13 +3,25 @@
...
+///
+/// To reference an attribute on a Slice entity or file, you can use '$attributes' after the source, followed by the
+/// index of the attribute to reference, and optionally, the index of an argument to that attribute. For example:
+/// 'MyMod::MyEnum::$attributes::0' references the first attribute of 'MyEnum', and '#MyFile.slice::$attributes::0::1'
+/// references the second argument of the first attribute on 'MyFile.slice'.
+typealias Source = string
```
externl: Could we use a type with a little more structure here? Maybe a variant enum.

```
enum Source {
    Identifier(id: string, attribute: tag(1) string)
    FilePath(path: string, attribute: tag(1) string) 
}
```

Or maybe something like:

```
struct Source {
   id: string
   sourceType: ABasicEnum
   attribute: tag(1) string
}
```

slice/Compiler/Diagnostics.slice
```diff
@@ -3,13 +3,25 @@
...
+
 /// A message that is reported to provide information and context for errors, warnings, or other issues.
 struct Diagnostic {
     /// The exact kind of diagnostic.
     kind: DiagnosticKind
 
     /// The fully-scoped Slice identifier of the element that this diagnostic is in reference to (if there is one).
-    source: string?
+    source: Source?
```
pepone: Not sure about this alias for string, Source seems pretty generic for this, and we define in the top-level compiler module.

Did you consider defining a separate type for Source, and avoid the string parsing, so far that is one of the advantages of using Slice structured data to communicate between the compiler and the plugins.

slicec/src/main.rs
```diff
@@ -75,21 +80,21 @@ fn spawn_plugin_process(plugin: &Plugin, slice_payload: &[u8]) -> std::io::Resul
...
-        // Obtain an exclusive handle to this process's 'stderr'.
-        let mut stderr = std::io::stderr().lock();
-
-        // Pipe the output from the subprocess's 'stderr' to this process's 'stderr', and then return.
-        let error = match stderr.write_all(&output.stderr) {
-            Ok(_) => Error::other("errors reported on 'stderr'"),
-            Err(err) => Error::new(ErrorKind::BrokenPipe, err),
-        };
-        return Err(error);
+        // TODO: switch to 'from_utf8_lossy_owned' when stabilized (https://github.com/rust-lang/rust/issues/129436).
```
pepone, line 94: better move to GitHub issue

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,237 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+    //    let message = format!("argument index '{argument_index}' is out of bounds");
+    //    let error = SlicecDiagnostic::from_error(SlicecError::Other { message })
+    //        .add_note(CRITICAL_FLAW_STRING, None);
+    //    output.push(error);
+    //    return None;
+    //};
+    //// Return the span of the argument.
+    //Some(argument.span().clone())
+
+    // TODO: we need to refactor the 'Attributes' API to expose arguments and their spans before we can implement this.
```
pepone, line 255: We should avoid adding the commented out code

slice/Compiler/Diagnostics.slice
InsertCreativityHere: We could, but I'd rather avoid using URI-like syntax, unless the thing is an actual URI, and here it isn't.
We support no query params, fragments, and will never properly escape this thing.
Also, it'd be weird to use a file-uri for files, but then just "MyMod::MyStruct" the other 80% of the time.

I'm open to replacements for '#', but I don't like using 'file://' for something which isn't actually a URI.

slicec/src/main.rs
InsertCreativityHere: Yes, I'll definitely open an issue, but there's no point opening it unless/until this PR is merged.
Right now, the code in question isn't on main.

InsertCreativityHere:
> Did you consider defining a separate type for Source, and avoid the string parsing

I only briefly considered it, but since you and @externl both asked, I spent some time re-implementing it with more structure to see the difference. And after seeing it side by side, it's not clear that the structure helps much.
Specifically I tried both (slightly modified) versions of the types in [this comment](https://github.com/icerpc/slicec/pull/776#discussion_r3259848940).

The main takeaway from this was:
- Creating strings is very simple and easy, creating typed structures is more unwieldy. The C# code is simpler with strings.
- Parsing strings is annoying, and using typed structures eliminates it. The Rust code is simpler with structured types.
- It is nicer though, to not have these 'magic' tokens like `$attributes`

But, most of the annoyance in the Rust code is actually from _validating_ the source, not _parsing_ it.
And, unexpectedly, the structured type requires _more validation_ than the string. It also just feels a little bloated with specific fields for `attributeIndex`, and `attributeArgumentIndex`. And we need to handle weird cases like what if you set `attributeArgumentIndex`, but not `attributeIndex`? These weird cases aren't possible with the string, which uses order to disambiguate.
So, the simplicity savings in Rust is not much. We save some on eliminating parsing, but gain some in terms of extra validation.

Additionally, the "pain of parsing" is only paid once, by `slicec`.
Whereas the extra verbosity of constructing a type vs a string is paid by every single plugin we ever write.

Also, using these typed structures introduces much more code. Around 350 lines all in all:
- extra code in the Slice definitions
- more checked in generated code
- handwritten wrapper in `ZeroC.Slice.Symbols` for domain separation

In my opinion, these extra types and extra code are not worth the slight conceptual savings of not having `$attributes` or `#` for file.

----

Taking a step back. All that we actually want, is a location in a file so we can report a text snippet. That's it.
If we really want to introduce a highly structured approach, I'd rather just send/transmit locations directly (somehow) instead of having all this extra structure, just so `slicec` can convert it into a form it can use, just to look up the corresponding location of something.

slice/Compiler/Diagnostics.slice
InsertCreativityHere: See https://github.com/icerpc/slicec/pull/776#issuecomment-4500405442

I tried using your proposed types, and I'm not convinced there's alot of savings.
Using `tag(1) attribute: string` doesn't eliminate much parsing, since we still have to parse this string.
And using `tag(1) attributeIndex: u8, tag(2) attributeArgumentIndex: u8` introduces edge cases we need to validate that counteract the savings we get from avoiding parsing.

Using `sourceKind: ABasicEnum` or making `Source` a variant enum (but otherwise not splitting the arguments into a separate string) is better, but still means that `Source` has to become a real type instead of a simple alias right now. This adds about 180 lines of code between C# and Rust (between the checked in generated code, and the domain wrapper type).
And all that it saves us is a `starts_with('#')` check.

slice/Compiler/Diagnostics.slice
InsertCreativityHere: Yes, you're right. I changed this to `DiagnosticSource`. Whether we keep an alias or change to a struct, enum etc. I think this is a better name.

For using a type vs just a string, see https://github.com/icerpc/slicec/pull/776/#issuecomment-4500405442
But since it's long: **TL;DR** the parsing isn't where most of the cruff comes from, it actually comes from validation and lookup, which don't care about the on-the-wire representation (like if a plugin reported attribute # 6 but there's only 4 attributes, you said "FooBar", but no element exists with that name, etc).

Switching to a structured approach saves us some validation (since we can get rid of `$attributes`), but introduces new validation in other ways (what if you pass an attribute argument index, but not an argument? (not possible with the string representation)). So, it ends up being about equal really.

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,237 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+                actual_count: actual_count as usize,
+            })
+        }
+
+        DiagnosticKind::Unknown { discriminant, fields_payload } => {
+            let mut message = format!("received an unknown diagnostic with a code of '{discriminant}'");
+            if !fields_payload.is_empty() {
+                write!(message, " and field-payload of:\n{fields_payload:?}").unwrap();
+            }
+            SlicecDiagnosticKind::Error(SlicecError::Other { message})
```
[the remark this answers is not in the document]
InsertCreativityHere: This is intentional and fine. A plugin reporting an error code that `slicec` isn't aware of is allowed by the spec (since we use an `unchecked enum`. Unfortunately, we can't match this to a well-known error, but this is definitely not a "critical flaw" in the same as returning `source: "UKYRI&RL^RV:(*&*^*&)"` is.

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,237 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+            => SlicecDiagnosticKind::Error(SlicecError::InvalidAttributeArgument { directive, argument }),
+
+        DiagnosticKind::IncorrectAttributeArgumentCount { directive, min_expected, max_expected, actual_count } => {
+            let min_expected = if min_expected == u8::MAX { usize::MAX } else { min_expected as usize };
+            let max_expected = if max_expected == u8::MAX { usize::MAX - 1 } else { max_expected as usize };
+            SlicecDiagnosticKind::Error(SlicecError::IncorrectAttributeArgumentCount {
+                directive,
+                expected_count: min_expected..(max_expected + 1),
+                actual_count: actual_count as usize,
+            })
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes, `max_expected == u8::MAX` is a special sentinel value meaning: this accepts up to ∞-many arguments.
For `min_expected`, indeed, an attribute which requires a minimum of ∞ arguments wouldn't be very useful.

Fixed.

slice/Compiler/Diagnostics.slice
```diff
@@ -3,13 +3,25 @@
...
+///
+/// The source of a diagnostic can be either:
+/// - A reference to a Slice entity, written as its fully-scoped identifier; ex: 'MyModule::MyInterface::myOp'.
+/// - A reference to a Slice file, written as that file's relative path prefixed with '#'; ex: '#dir1/dir2/file.slice'.
+///
+/// To reference an attribute on a Slice entity or file, you can use '$attributes' after the source, followed by the
+/// index of the attribute to reference, and optionally, the index of an argument to that attribute. For example:
+/// 'MyMod::MyEnum::$attributes::0' references the first attribute of 'MyEnum', and '#MyFile.slice::$attributes::0::1'
+/// references the second argument of the first attribute on 'MyFile.slice'.
+typealias DiagnosticSource = string
```
pepone, line 16: How do you reference an attribute for an operation parameter, or return, does this work?

slice/Compiler/Diagnostics.slice
```diff
@@ -3,13 +3,25 @@
 [cs::identifier("ZeroC.Slice.Symbols.Compiler")]
 module Compiler
 
+/// Represents the source of a {@link Diagnostic}, which can be used to provide context and snippets during emission.
+///
+/// The source of a diagnostic can be either:
+/// - A reference to a Slice entity, written as its fully-scoped identifier; ex: 'MyModule::MyInterface::myOp'.
```
pepone, line 9: Can you give a example where a pluhin might want to emit an error for a identifier? 

In my view the generator would only validate attributes which can be specific to a plug-in, other diagnostics are handled by slicec and are plugin agnostic.

I guess if we can need it if we end-up not supporting some Slice features in a generator.

slicec/src/slice_file_converter.rs
```diff
@@ -29,9 +29,237 @@ use slicec::grammar::{Attributable, Commentable, Contained, Entity, Member, Name
...
+                    .push_into(output);
+                return (None, None);
+            }
+        };
+
+        attributable = slice_file;
+        // The 'scope' is always `None`, since files do not logically have a Slice scope like symbols do.
+        scope = None;
+        // If the diagnostic source is the file itself, we use an empty span that points to the start of the file.
+        default_span = Some(Span::new((1, 1).into(), (1, 1).into(), &slice_file.relative_path));
```
pepone, line 157: What is an example of a file diagnostic a plugin will emit?

pepone (approved):

externl (approved):

slice/Compiler/Diagnostics.slice
InsertCreativityHere: Yes, this already works. This doesn't have to be a `type ID` like we have in old Slice.
The comment states this is an `entity ID`, which isn't restricted to just types.

For example, if you wanted to report the 4th attribute of a parameter, you'd do:
`MyModule::MyInterface::MyOperation::Parameter3::$attributes::4`

This also works for return values, with one special case.
If you have a single return type, `slicec` will implicitly give it the name `returnValue`.
But this doesn't matter to the code-generator. All it needs to do is call `thing.Identifier`, and it will get the correct values.

slice/Compiler/Diagnostics.slice
InsertCreativityHere: > I guess if we can need it if we end-up not supporting some Slice features in a generator.

Yeah, this is more about future proofing. But, I think there is already one possible use case.
I think it'd be good for the C# code-gen to report an `info` diagnostic saying its skipping generating code for `typealias FooBar`.

Typealias is a first class Slice feature, and we'll generate code for it in Rust, Swift, and Python.
We don't generate code for it in C#. Would be good to explicitly report this IMO.

slicec/src/slice_file_converter.rs
InsertCreativityHere: Well, at the moment it could report that the generator wants to write twice to the same file, although I want to move this check into `slicec`.

Other languages could also report file-level attributes. We don't have any C# specific ones anymore (since we've replaced `cs::namespace` with `cs::identifier`), but this is just a coincidence. Language specific attributes are perfectly possible right now, and will be necessary to properly support `rs::attribute` in Rust.

Issue #777, opened 2026-05-12
Filter Out Duplicate Diagnostic Messages


InsertCreativityHere:
`slicec` should check for, and filter out, duplicate diagnostics before emitting them.

This is actually going to be a fairly common occurrence, since most of our language plugins are split into 2 plugins, which share validation logic (meaning the validation logic will be run twice, generating 2 diagnostics for each offense).

InsertCreativityHere:
This was implemented in #778.

PR #778, opened 2026-05-18
Don't Emit Duplicate Diagnostics (& Fix Derive Order)

InsertCreativityHere:
This PR implements #777.

Now that I've gotten #776 working locally, I've seen that every diagnostic the generator reports is duplicated (because the `ZeroC` generator and the `IceRpc` generator both do their own validation). I don't think there's any good way to avoid this double validation, and I think it's good to keep the generators separate.

So, this PR changes `slicec` to detect and suppress duplicate diagnostics, leaving plugins free to not worry about this.
We do the de-duplication at the point where we convert all our diagnostics into their 'annotated' form (the form we actually print to the user). This way `slicec` is free to report and collect as many `Diagnostic`s as it wants, and then at the very end, right before we emit them, is when we filter.

### DIFF
--- a/slicec/src/compilation_state.rs
+++ b/slicec/src/compilation_state.rs
@@ -42,10 +42,21 @@ impl CompilationState {
         }
     }
 
+    /// Creates an [`AnnotatedDiagnostic`] for each (`Diagnostic`)[crate::diagnostics::Diagnostic] stored in this
+    /// `CompilationState`. This does not "use up" the diagnostics. Calling this function multiple times will yield the
+    /// same output.
+    ///
+    /// The returned diagnostics are also de-duplicated. Duplicates are expected when slicec calls multiple plugins
+    /// which may run identical validation. If multiple diagnostics are _exact_ duplicates, only the first will be
+    /// present in the returned [`Vec`].
     pub fn get_annotated_diagnostics(&self, options: &SliceOptions) -> Vec<AnnotatedDiagnostic> {
-        self.diagnostics
-            .iter()
-            .map(|diagnostic| crate::diagnostics::convert_diagnostic(diagnostic, options, self))
-            .collect()
+        let mut annotated_diagnostics = Vec::new();
+        for diagnostic in &*self.diagnostics {
+            let converted = crate::diagnostics::convert_diagnostic(diagnostic, options, self);
+            if !annotated_diagnostics.contains(&converted) {
+                annotated_diagnostics.push(converted);
+            }
+        }
+        annotated_diagnostics
     }
 }
--- a/slicec/src/compilation_state.rs
+++ b/slicec/src/compilation_state.rs
@@ -42,7 +42,7 @@ impl CompilationState {
         }
     }
 
-    /// Creates an [`AnnotatedDiagnostic`] for each (`Diagnostic`)[crate::diagnostics::Diagnostic] stored in this
+    /// Creates an [`AnnotatedDiagnostic`] for each [`Diagnostic`](crate::diagnostics::Diagnostic) stored in this
     /// `CompilationState`. This does not "use up" the diagnostics. Calling this function multiple times will yield the
     /// same output.
     ///
--- a/slicec/src/diagnostics/annotated_diagnostic.rs
+++ b/slicec/src/diagnostics/annotated_diagnostic.rs
@@ -9,7 +9,7 @@ use serde::Serialize;
 
 /// An annotated version of a [`Diagnostic`], whose [`DiagnosticLevel`] has been computed (taking into account any
 /// 'allow' attributes or command-line flags), and that has pre-extracted text snippets to display alongside messages.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct AnnotatedDiagnostic {
     pub message: String,
     pub level: DiagnosticLevel,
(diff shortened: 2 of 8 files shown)


slicec/src/diagnostics/annotated_diagnostic.rs
```diff
@@ -18,13 +18,13 @@ pub struct AnnotatedDiagnostic {
     pub notes: Vec<AnnotatedNote>,
 }
 
-#[derive(Debug, Clone, Serialize)]
+#[derive(Serialize, Clone, Debug, PartialEq, Eq)]
```
externl: Does Rust have a preferred ordering for these?

externl (approved):

slicec/src/diagnostics/annotated_diagnostic.rs
```diff
@@ -9,7 +9,7 @@ use serde::Serialize;
 
 /// An annotated version of a [`Diagnostic`], whose [`DiagnosticLevel`] has been computed (taking into account any
 /// 'allow' attributes or command-line flags), and that has pre-extracted text snippets to display alongside messages.
```
InsertCreativityHere, line 11: Re-arranged the derives to be in proper order.
Any non-standard traits should come first (`Serialize` is a 3rd party trait here), followed by any standard-library traits in alphabetical order (except for `PartialEq`, and `PartialOrd` which should come right after their corresponding `Eq` and `Ord` respectively).

slicec/src/compilation_state.rs
```diff
@@ -42,10 +42,21 @@ impl CompilationState {
...
     }
 
+    /// Creates an [`AnnotatedDiagnostic`] for each (`Diagnostic`)[crate::diagnostics::Diagnostic] stored in this
+    /// `CompilationState`. This does not "use up" the diagnostics. Calling this function multiple times will yield the
+    /// same output.
+    ///
+    /// The returned diagnostics are also de-duplicated. Duplicates are expected when slicec calls multiple plugins
+    /// which may run identical validation. If multiple diagnostics are _exact_ duplicates, only the first will be
+    /// present in the returned [`Vec`].
     pub fn get_annotated_diagnostics(&self, options: &SliceOptions) -> Vec<AnnotatedDiagnostic> {
```
InsertCreativityHere, line 52: The actual filtering.

slicec/src/diagnostics/annotated_diagnostic.rs
InsertCreativityHere: You were too fast '~'
I wish GitHub would let you pre-comment on a PR before opening it...

slicec/tests/diagnostic_output_tests.rs
```diff
@@ -109,8 +113,62 @@ error [E008]: invalid enum 'E': enums must contain at least one enumerator
...
+            diagnostic2_1,
+            diagnostic3_1,
+            diagnostic4_1,
+            diagnostic1_3,
+            diagnostic3_2,
+        ]);
+
+        // Act
+        let converted_diagnostics = state.get_annotated_diagnostics(&SliceOptions::default());
+        println!("{converted_diagnostics:?}");
```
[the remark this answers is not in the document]
InsertCreativityHere: Leftover from me double-checking stuff! Removed.

slicec/src/compilation_state.rs
```diff
@@ -42,10 +42,21 @@ impl CompilationState {
...
-            .iter()
-            .map(|diagnostic| crate::diagnostics::convert_diagnostic(diagnostic, options, self))
-            .collect()
+        let mut annotated_diagnostics = Vec::new();
+        for diagnostic in &*self.diagnostics {
+            let converted = crate::diagnostics::convert_diagnostic(diagnostic, options, self);
+            if !annotated_diagnostics.contains(&converted) {
+                annotated_diagnostics.push(converted);
+            }
+        }
```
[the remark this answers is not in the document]
InsertCreativityHere: I don't think there is a better way. We want to preserve order: diagnostics should be emitted in the order that they were reported. Also, it'd feel weird for `AnnotatedDiagnostic` to implement `Hash`. These aren't things that really make alot of sense to hash...

I also think it's fine to not over-optimize for the case where a user has 1,000 errors in their Slice files.

slicec/src/compilation_state.rs
```diff
@@ -42,10 +42,21 @@ impl CompilationState {
         }
     }
 
+    /// Creates an [`AnnotatedDiagnostic`] for each (`Diagnostic`)[crate::diagnostics::Diagnostic] stored in this
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed.

slicec/src/diagnostics/annotated_diagnostic.rs
InsertCreativityHere: Had to add `Eq` and `PartialEq` so that `AnnotatedDiagnostic`s are comparable to one another.

`derive(Eq)` generates an _exact_ equality. Every single field must be exactly identical for the whole thing to be equal.

slicec/tests/diagnostic_output_tests.rs
```diff
@@ -4,11 +4,14 @@ mod test_helpers;
 
 mod output {
     use crate::test_helpers::parse;
+    use slicec::compilation_state::CompilationState;
     use slicec::diagnostic_emitter::DiagnosticEmitter;
+    use slicec::diagnostics::{Diagnostic, Error, Lint};
     use slicec::slice_options::{DiagnosticFormat, SliceOptions};
 
     #[test]
     fn output_to_json() {
```
InsertCreativityHere, line 13: I also added `arrange` comments to this test file.
I don't know why, but this file was missing them.

pepone (approved):

Issue #720, opened 2024-04-24
Behavior for Duplicate Dictionary Entries?


InsertCreativityHere:
Like many languages, Rust's main dictionary type is the `HashMap`.
An unordered map, with unique keys.

What should the behavior be for decoding a dictionary with duplicate keys?
For example, if this is received across the wire: `[3][(5, "hi"), (12, "he"), (5, "ha")]`.

The language guide is very clear on this. One of the first points we state in the dictionary docs is that each key must be unique:
https://docs.icerpc.dev/slice2/language-guide/dictionary-types

And, in C#, we strictly enforce the spec, but I suspect we only do so by accident.
[`Dictionary.Add`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.idictionary-2.add?view=net-8.0) throws an exception if you attempt to add a key which already exists.

Should we also strictly enforce the spec in Rust too?
By returning a decoding error if we encounter a duplicate key.
Currently, we just call [`HashMap::insert`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert), which will happily update an existing key with a new value upon seeing a duplicate.

Some nuance:
The receiver still upholds the rule. Our `HashMap` _does_ still have unique keys. The real problem is with the sender.
The sender sent us a 'bad' dictionary. So, is it really worth explicitly checking each insertion for uniqueness when:
- this only happens in the rare case where I was sent garbage, which is on them.
- and whether or not we return an error, we will still be upholding all the requirements of the type?

InsertCreativityHere:
Bernard's suggestion for icerpc-csharp was implemented in https://github.com/icerpc/icerpc-csharp/pull/4518.

InsertCreativityHere:
This was finally finished in #780.

PR #780, opened 2026-05-19
Reject Duplicate Dictionary Keys

InsertCreativityHere:
This PR implements #720 by strongly enforcing the spec, i.e. if we encounter duplicate keys during decoding, we error.

I don't see value in mirroring this check on the encoding side. If you're using any normal dictionary type in Rust, they will already guarantee uniqueness of keys. And if you're using your own type, well, there's nothing we can do to stop you from sending duplicates anyways.

### DIFF
--- a/slice-codec/src/decode_from.rs
+++ b/slice-codec/src/decode_from.rs
@@ -56,7 +56,8 @@ macro_rules! decode_dictionary_entries {
                 return Err(error.into());
             } else {
                 // `insert` should always return `None` for a new key.
-                debug_assert!($dictionary.insert(key, value).is_none());
+                let result = $dictionary.insert(key, value);
+                debug_assert!(result.is_none());
             }
         }
     };
--- a/slice-codec/src/decode_from.rs
+++ b/slice-codec/src/decode_from.rs
@@ -54,11 +54,10 @@ macro_rules! decode_dictionary_entries {
                     key: alloc::format!("{key:?}"),
                 };
                 return Err(error.into());
-            } else {
-                // `insert` should always return `None` for a new key.
-                let result = $dictionary.insert(key, value);
-                debug_assert!(result.is_none());
             }
+
+            let result = $dictionary.insert(key, value);
+            debug_assert!(result.is_none()); // `insert` should always return `None` for a new key.
         }
     };
 }
--- a/slice-codec/src/decode_from.rs
+++ b/slice-codec/src/decode_from.rs
@@ -41,16 +41,25 @@ pub use implement_decode_from_on_numeric_primitive_type; // Re-export the macro 
 #[doc(hidden)]
 #[macro_export]
 macro_rules! decode_dictionary_entries {
-    ($map:ident, $decoder:ident, $length:ident) => {
-        // Decode each entry, and insert them into the map, one by one.
+    ($dictionary:ident, $decoder:ident, $length:ident) => {
+        // Decode each entry, and insert them into the dictionary, one by one.
         for _ in 0..$length {
             let key = $decoder.decode()?;
             let value = $decoder.decode()?;
-            if let Some(_duplicate) = $map.insert(key, value) {
-                // TODO
-                // If you insert a duplicate key into the map, it will return the old key.
-                // So, if we hit this, we return  an error, because dictionary keys must be unique.
-                todo!();
+
+            // Check if the key already exists in the dictionary
+            match $dictionary.contains_key(&key) {
+                false => {
+                    // If it doesn't, insert it normally, and assert that 'insert' returned `None`.
+                    debug_assert!($dictionary.insert(key, value).is_none());
+                }
+                true => {
+                    // If it does, immediately return a duplicate key error.
+                    let error = InvalidDataErrorKind::DuplicateDictionaryKey {
+                        key: alloc::format!("{key:?}"),
+                    };
+                    return Err(error.into());
+                }
             }
         }
     };
--- a/slice-codec/src/decode_from.rs
+++ b/slice-codec/src/decode_from.rs
@@ -47,7 +47,7 @@ macro_rules! decode_dictionary_entries {
             let key = $decoder.decode()?;
             let value = $decoder.decode()?;
 
-            // Insert the key,value pair into the dictionary. If the decoded key is already present, we return an error.
+            // Insert the (key,value) pair into the dictionary. If the decoded key is already present, we return an error.
             if $dictionary.insert(key, value).is_some() {
                 return Err(InvalidDataErrorKind::DuplicateDictionaryKey.into());
             }
--- a/slice-codec/src/decode_from.rs
+++ b/slice-codec/src/decode_from.rs
@@ -47,7 +47,7 @@ macro_rules! decode_dictionary_entries {
             let key = $decoder.decode()?;
             let value = $decoder.decode()?;
 
-            // Insert the (key,value) pair into the dictionary. If the decoded key is already present, we return an error.
+            // Insert the (key,value) pair into the dictionary. If the decoded key is already present, return an error.
             if $dictionary.insert(key, value).is_some() {
                 return Err(InvalidDataErrorKind::DuplicateDictionaryKey.into());
             }
--- a/slice-codec/src/decode_from.rs
+++ b/slice-codec/src/decode_from.rs
@@ -47,19 +47,16 @@ macro_rules! decode_dictionary_entries {
             let key = $decoder.decode()?;
             let value = $decoder.decode()?;
 
-            // Check if the key already exists in the dictionary
-            match $dictionary.contains_key(&key) {
-                false => {
-                    // If it doesn't, insert it normally, and assert that 'insert' returned `None`.
-                    debug_assert!($dictionary.insert(key, value).is_none());
-                }
-                true => {
-                    // If it does, immediately return a duplicate key error.
-                    let error = InvalidDataErrorKind::DuplicateDictionaryKey {
-                        key: alloc::format!("{key:?}"),
-                    };
-                    return Err(error.into());
-                }
+            // Check if the key already exists in the dictionary. If it does, we return a duplicate-key error.
+            // Otherwise, we insert the key normally.
+            if $dictionary.contains_key(&key) {
+                let error = InvalidDataErrorKind::DuplicateDictionaryKey {
+                    key: alloc::format!("{key:?}"),
+                };
+                return Err(error.into());
+            } else {
+                // `insert` should always return `None` for a new key.
+                debug_assert!($dictionary.insert(key, value).is_none());
             }
         }
     };
--- a/slice-codec/src/decode_from.rs
+++ b/slice-codec/src/decode_from.rs
@@ -47,17 +47,10 @@ macro_rules! decode_dictionary_entries {
             let key = $decoder.decode()?;
             let value = $decoder.decode()?;
 
-            // Check if the key already exists in the dictionary. If it does, we return a duplicate-key error.
-            // Otherwise, we insert the key normally.
-            if $dictionary.contains_key(&key) {
-                let error = InvalidDataErrorKind::DuplicateDictionaryKey {
-                    key: alloc::format!("{key:?}"),
-                };
-                return Err(error.into());
+            // Insert the key,value pair into the dictionary. If the decoded key is already present, we return an error.
+            if $dictionary.insert(key, value).is_some() {
+                return Err(InvalidDataErrorKind::DuplicateDictionaryKey.into());
             }
-
-            let result = $dictionary.insert(key, value);
-            debug_assert!(result.is_none()); // `insert` should always return `None` for a new key.
         }
     };
 }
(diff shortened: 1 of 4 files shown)


slice-codec/src/decoding.rs
```diff
@@ -237,7 +239,7 @@ where
 #[cfg(feature = "std")]
 impl<K, V> DecodeFrom for HashMap<K, V>
 where
-    K: DecodeFrom + Eq + Hash,
+    K: DecodeFrom + Debug + Eq + Hash,
```
InsertCreativityHere, line 242: The `Debug` trait means "you can print a low-level string representation of this type".
We need this in order to print the duplicate key in the error message.

And adding this trait bound costs nothing, since all the key types we care about implement `Debug`.

slice-codec/src/decode_from.rs
```diff
@@ -41,16 +41,25 @@ pub use implement_decode_from_on_numeric_primitive_type; // Re-export the macro
...
-                // TODO
-                // If you insert a duplicate key into the map, it will return the old key.
-                // So, if we hit this, we return  an error, because dictionary keys must be unique.
-                todo!();
+
+            // Check if the key already exists in the dictionary
+            match $dictionary.contains_key(&key) {
+                false => {
+                    // If it doesn't, insert it normally, and assert that 'insert' returned `None`.
+                    debug_assert!($dictionary.insert(key, value).is_none());
```
InsertCreativityHere: `insert` should always return `None` if the key is unique.

slice-codec/src/error.rs
```diff
@@ -182,13 +189,18 @@ impl Display for InvalidDataErrorKind {
...
+            #[cfg(feature = "alloc")]
+            Self::InvalidString(inner) => inner.fmt(f),
+
             Self::OutOfRange { value, min, max, typename } => {
                 write!(
                     f,
                     "value '{value}' is outside the allowed range for type '{typename}'; values must be within [{min}..{max}]"
                 )
             }
-            _ => todo!(),
```
InsertCreativityHere, line 191: While touching this code I also implemented this `todo!()`.
Seems like a reminder to provide a message for `InvalidString` errors (if we decode invalid utf8).

slice-codec/src/decode_from.rs
```diff
@@ -41,16 +41,22 @@ pub use implement_decode_from_on_numeric_primitive_type; // Re-export the macro
...
-                // If you insert a duplicate key into the map, it will return the old key.
-                // So, if we hit this, we return  an error, because dictionary keys must be unique.
-                todo!();
+
+            // Check if the key already exists in the dictionary. If it does, we return a duplicate-key error.
+            // Otherwise, we insert the key normally.
+            if $dictionary.contains_key(&key) {
+                let error = InvalidDataErrorKind::DuplicateDictionaryKey {
+                    key: alloc::format!("{key:?}"),
+                };
```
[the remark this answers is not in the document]
InsertCreativityHere: This is fine for a number of reasons.
A malformed key is a much more serious error than a duplicate key, and so it makes sense to "give preference" to this error. Also, it seems cleaning to do all the decoding up front, then do the validation, instead of a piecemeal approach. Doing slightly more work in an unlikely error condition is fine.

slice-codec/src/decode_from.rs
```diff
@@ -41,16 +41,22 @@ pub use implement_decode_from_on_numeric_primitive_type; // Re-export the macro
...
+                return Err(error.into());
+            } else {
+                // `insert` should always return `None` for a new key.
+                debug_assert!($dictionary.insert(key, value).is_none());
```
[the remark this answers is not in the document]
InsertCreativityHere: This doesn't work, although I wish that it did.
This was the original implementation I tried to get working, but doing it this way makes it impossible to report what the bad key was. And I feel like reporting the duplicate key value is worthwhile enough to be worth the extra call.

I also suspect that the optimizer is smart enough to see that the 2nd check if unnecessary and can be elided. Who knows though.

slice-codec/src/decoding.rs
```diff
@@ -237,7 +239,7 @@ where
...
     V: DecodeFrom,
 {
```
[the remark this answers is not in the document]
InsertCreativityHere: We already require that the key implements `Eq` and `Hash`, which are non-trivial bounds to support. `Debug` is literally the most common and simplest trait to support. Adding it as a bound is fine IMO.

slice-codec/src/error.rs
```diff
@@ -170,6 +170,13 @@ pub enum InvalidDataErrorKind {
         max: i128,
         typename: &'static str,
     },
+
+    /// A key appears multiple times in a dictionary, violating the uniqueness requirement.
+    #[cfg(feature = "alloc")]
+    DuplicateDictionaryKey {
+        /// A string representation of the offending key.
+        key: String,
+    },
```
InsertCreativityHere: The new error kind so we can print a nice and specific message for it.

It is standard in Rust to have an incredibly granular `ErrorKind` enum like this. Although it's discouraged for an end user to go digging through the individual variants.

See `io::Error` in `std`: https://doc.rust-lang.org/std/io/enum.ErrorKind.html or `DecodeErrorKind` in `prost`: https://docs.rs/prost/latest/src/prost/error.rs.html#106-137 etc.

slice-codec/src/decode_from.rs
```diff
@@ -41,16 +41,22 @@ pub use implement_decode_from_on_numeric_primitive_type; // Re-export the macro
...
             }
```
externl: In other languages it's generally bad practice to perform any operations inside a debug statement. Is the same true for rust?

InsertCreativityHere: Yes, you're right. I moved it outside of the assert.
Let me check if this causes any unused variable warnings in release mode though...

slice-codec/src/decode_from.rs
```diff
@@ -56,7 +56,8 @@ macro_rules! decode_dictionary_entries {
                 return Err(error.into());
```
externl: For terminal if statements we should not use the else.

externl (approved):

slice-codec/src/decode_from.rs
InsertCreativityHere: Fair enough, I removed the `else`.

The if-else was really just leftover from an older approach I tried first.

slice-codec/src/decode_from.rs
```diff
@@ -41,17 +41,23 @@ pub use implement_decode_from_on_numeric_primitive_type; // Re-export the macro
...
         for _ in 0..$length {
             let key = $decoder.decode()?;
             let value = $decoder.decode()?;
-            if let Some(_duplicate) = $map.insert(key, value) {
-                // TODO
-                // If you insert a duplicate key into the map, it will return the old key.
-                // So, if we hit this, we return  an error, because dictionary keys must be unique.
-                todo!();
+
+            // Check if the key already exists in the dictionary. If it does, we return a duplicate-key error.
```
bernardnormier: We should not optimize for the "bad case" and make decoding slower.

Here, we should insert then check the result.

slice-codec/src/decode_from.rs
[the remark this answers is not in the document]
bernardnormier: I am fine with decoding key and value first before the insertion.

slice-codec/src/decode_from.rs
```diff
@@ -41,17 +41,23 @@ pub use implement_decode_from_on_numeric_primitive_type; // Re-export the macro
...
+            // Otherwise, we insert the key normally.
+            if $dictionary.contains_key(&key) {
+                let error = InvalidDataErrorKind::DuplicateDictionaryKey {
+                    key: alloc::format!("{key:?}"),
+                };
+                return Err(error.into());
             }
+
+            let result = $dictionary.insert(key, value);
+            debug_assert!(result.is_none()); // `insert` should always return `None` for a new key.
```
bernardnormier: That's what we should check. No call to contains_key.

slice-codec/src/decode_from.rs
InsertCreativityHere: This isn't done for optimization's sake. My original implementation was exactly what you say: call insert then check.

The issue is that `insert` moves the key. Even if there's a duplicate, once you call `insert` the key is gone.
Meaning there's no way to say which key was duplicated in the error message. Which I want to do, and is what we do in C#.

There is [`try_insert`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.try_insert) which does exactly what we want! But, it's `nightly` only, so we can't use it. So we either:
1) check `contains` before `insert`, letting us report the bad key.
2) call `insert` and check the result, but we can't report the bad key.
3) Use the `Entry` API which avoids doing a double-lookup, but, this might introduce it's own performance overhead. Need to look into it.

I chose option 1 because it gives better error messages, and because I suspect the optimizer can avoid this double-checking in practice. I'll look at the IR that gets spit out and see if this is the case or not though.

InsertCreativityHere:
I did some benchmarking of 3 different approaches:
1) Inserting the key without checking if it exists first (causing a subpar error message)
2) This PR's current implementation (check `contains` before `insert`ing)
3) Using the `Entry` API which avoids a double check, but requires stack-allocations.

The result is that: **all 3 of these are basically the same**, except for decoding small dictionaries of primitive types.
For a 100-element long dictionary, not checking if the key exists saves around `0.2ns` per element.
But even bumping the element count up to 1000 completely erases this difference (in fact at 1000 elements, checking `contains` first was _faster_. This is obviously just a random fluctuation though, showing how close the timings must be).

I think that providing a very good error message is worth losing `0.2ns` in the one case where you're sending 100 or fewer dictionaries containing only primitive types. Also, this performance hit is only temporary. Once [`try_insert`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.try_insert) is stabilized, we will switch to that, which will eliminate the need to call `contains_key` first.

# Numbers

(These numbers are just a summary, if you want the full numbers, let me know, and I can send them)
Benchmarking was done through the Rust Criterion crate (the standard benchmarking tool in Rust).
Timing measurements are of the form `[MINIMUM_TIME, MEDIAN_TIME, MAXIMUM_TIME]`.

### Decoding 100 `<String, String>` entries:
- `direct_insert`: 300 measurements, `[6.5012 µs 6.7796 µs 7.2741 µs]` (6 mild outliers, 8 severe outliers)
- `contains_first`: 300 measurements, `[6.4684 µs 6.8026 µs 7.4078 µs]` (9 mild outliers, 8 severe outliers)
- `entry_api`: 300 measurements, `[6.6041 µs 6.8519 µs 7.1902 µs]` (8 mild outliers, 9 severe outliers)


### Decoding 100,000 `<String, String>` entries:
- `direct_insert`: 300 measurements, `[7.3492 ms 7.4752 ms 7.6740 ms]` (9 mild outliers, 2 severe outliers)
- `contains_first`: 300 measurements, `[7.3663 ms 7.4544 ms 7.5671 ms]` (13 mild outliers, 2 severe outliers)
- `entry_api`: 300 measurements, `[7.3148 ms 7.4722 ms 7.7285 ms]` (8 mild outliers, 7 severe outliers)


### Decoding 100 `<int32, varuint62>` entries:
- `direct_insert`: 300 measurements, `[142.68 ns 145.11 ns 149.87 ns]` (9 mild outliers, 3 severe outliers)
- `contains_first`: 300 measurements, `[162.34 ns 165.73 ns 171.18 ns]` (14 mild outliers, 7 severe outliers)
- `entry_api`: 300 measurements, `[165.18 ns 165.97 ns 166.86 ns]` (13 mild outliers, 7 severe outliers)


### Decoding 100,00 `<int32, varuint62>` entries:
- `direct_insert`: 200 measurements, `[39.582 µs 40.3115 µs 41.067 µs]` (8 mild outliers, 7 severe outliers, 5 below)
- `contains_first`: 300 measurements, `[38.965 µs 39.539 µs 40.030 µs]` (7 mild outliers, 10 severe outliers, 4 below)
- `entry_api`: 200 measurements, `[39.998 µs 40.429 40.914 µs]` (9 mild outliers, 4 severe outliers, 5 below)

slice-codec/src/decode_from.rs
InsertCreativityHere: This PR's approach is the best way to insert, while still being able to report the bad key.
The performance hit is truly miniscule (see https://github.com/icerpc/slicec/pull/780#issuecomment-4511067813), and only temporary. Once `try_insert` is stabilized, we can switch to it and even this small performance hit will vanish.

slice-codec/src/decode_from.rs
bernardnormier: We should do "2". Not reporting which key is duplicated is not a must-have. It's an unrecoverable decoding error which means the sender did something wrong.

slice-codec/src/decoding.rs
InsertCreativityHere: Even if we don't use it yet, this `Debug` bound will let us print the bad key in the future.
And every key type we care about implements `Debug`, so it's not a constraining bound.

bernardnormier (approved):

Issue #762, opened 2026-04-07
Add Check to Detect Overwritten Files


InsertCreativityHere:
We should add a check to `slicec` which keeps track of which paths it's written generated code to.
And if a single path would be written to more than once (in a single compilation run), we should issue a warning that the file was overwritten.

This could happen for a variety of reasons, and it's simpler for `slicec` to handle it, then asking each code-generator to include this check.

InsertCreativityHere:
This was implemented by #783 and https://github.com/icerpc/icerpc-csharp/pull/4690

PR #783, opened 2026-05-26
Reject Writing to the Same Path Multiple Times in a Single Compilation Run

InsertCreativityHere:
Implements #762.

----

We currently have a check in the C# code-generator to report an error if 2 `GeneratedFile`s have the same path.
This current check is not ideal though since:
- it would have to be implemented by every single generator
- it misses the more interesting case of cross-plugin overwriting

So, I've added this check to `slicec` itself. It keeps a map of written-to-paths and reports errors upon repeats.
If merged, I will remove the (now redundant) check from C#.

----

An easy way to test is to compile 2 files: `Test.slice` and `Test.IceRpc.slice`.
This causes a collision, and the following message (with this PR):
```
error [E000]: the path 'Test.IceRpc.cs' was already written to by '.\icerpc-csharp\src\ZeroC.Slice.Generator\bin\Debug\net10.0\slicec-csharp-generator.bat', and would be overwritten by '.\icerpc-csharp\src\IceRpc.Slice.Generator\bin\Debug\net10.0\slicec-icerpc-csharp-generator.bat'
```

### DIFF
--- a/slicec/src/main.rs
+++ b/slicec/src/main.rs
@@ -164,6 +164,30 @@ fn write_generated_file(
     Ok(())
 }
 
+/// Checks if the path of the provided `generated_file` has already been written to during this compilation.
+/// If it has, this returns an `Err` with an error describing the overwriting. Otherwise this returns `Ok`.
+/// Note that this function normalizes/canonicalizes the path before checking.
+fn check_if_file_is_overwritten<'a>(
+    generated_file: &definition_types::GeneratedFile,
+    generator: &'a Plugin,
+    written_to_paths: &mut HashMap<PathBuf, &'a Plugin>,
+) -> Result<(), slicec::diagnostics::Error> {
+    // Attempt to canonicalize the path. If that fails for any reason, fallback to using the path string as-is.
+    let canonical_path = std::fs::canonicalize(&generated_file.path)
+        .unwrap_or_else(|_| generated_file.path.to_owned().into());
+
+    // If we've already written to the file's path in this compilation run, return an error.
+    if let Some(other_plugin) = written_to_paths.insert(canonical_path, generator) {
+        let message = format!(
+            "the path '{}' was already written to by '{}', and would be overwritten by '{}'",
+            &generated_file.path, &other_plugin.path, &generator.path,
+        );
+        Err(slicec::diagnostics::Error::Other { message })
+    } else {
+        Ok(())
+    }
+}
+
 /// Converts an [`std::io::Error`] that occurred while trying to write a generated file into a
 /// [`Diagnostic`](slicec::diagnostics::Diagnostic), and pushes it into the provided [`Diagnostics`] collection.
 fn report_file_writing_error(file_path: &String, io_error: std::io::Error, diagnostics: &mut Diagnostics) {
@@ -220,7 +244,7 @@ fn main() -> ExitCode {
 
         // Store which file-paths we've written code to, so we can reject overwrites.
         // Keys are file-paths, and values are the plugins that wrote to the path (for more actionable diagnostics).
-        let mut written_to_paths: HashMap<String, &Plugin> = HashMap::new();
+        let mut written_to_paths: HashMap<PathBuf, &Plugin> = HashMap::new();
 
         // Block on each generator process until they're finished. If a generator completes successfully,
         // we get the response payload from it, write any generated files in the payload, and store any diagnostics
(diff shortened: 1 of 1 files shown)


slicec/src/main.rs
```diff
@@ -232,8 +237,19 @@ fn main() -> ExitCode {
...
                 for generated_file in &generated_files {
+                    // If we've already written to the file's path in this compilation run, report an error.
+                    if let Some(original_plugin) = written_to_paths.insert(generated_file.path.clone(), generator) {
+                        let message = format!(
+                            "the path '{}' was already written to by '{}', and would be overwritten by '{}'",
+                            &generated_file.path, &original_plugin.path, &generator.path,
+                        );
+                        let overwrite_error = slicec::diagnostics::Error::Other { message };
+                        slicec::diagnostics::Diagnostic::from_error(overwrite_error).push_into(diagnostics);
+                    }
```
InsertCreativityHere, line 267: We just report the error but otherwise do nothing else different.
We're already at the very end of compilation, no use in early-exiting now.
And overwriting the file is simpler than skipping it.

Note that the fact we emit an error means the compilation will still fail with a non-zero exit code tho.

slicec/src/main.rs
```diff
@@ -232,8 +237,19 @@ fn main() -> ExitCode {
...
+
+                    // Write the generated file to disk.
+                    // If an error occurs while writing the file, report the error as a slicec diagnostic.
                     if let Err(io_error) = write_generated_file(generated_file, &slice_options.output_dir) {
-                        // If an error occurred while writing the file, report the error as a slicec diagnostic.
                         report_file_writing_error(&generated_file.path, io_error, &mut generator_diagnostics);
```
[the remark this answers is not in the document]
InsertCreativityHere: It's fine. This is a very rare error and neither file is "more correct" than the other.
It's simpler to just write the file, instead of adding a 2nd code path where we skip over it.

slicec/src/main.rs
```diff
@@ -232,8 +237,19 @@ fn main() -> ExitCode {
...
             if !generator_diagnostics.has_errors() {
                 for generated_file in &generated_files {
+                    // If we've already written to the file's path in this compilation run, report an error.
+                    if let Some(original_plugin) = written_to_paths.insert(generated_file.path.clone(), generator) {
+                        let message = format!(
+                            "the path '{}' was already written to by '{}', and would be overwritten by '{}'",
+                            &generated_file.path, &original_plugin.path, &generator.path,
+                        );
+                        let overwrite_error = slicec::diagnostics::Error::Other { message };
+                        slicec::diagnostics::Diagnostic::from_error(overwrite_error).push_into(diagnostics);
```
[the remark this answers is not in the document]
InsertCreativityHere: I think that the odds of 3 separate plugins all writing to the same file will never happen.
When `rust` stabilizes `try_insert`, we can switch to that instead of `insert`.

But I'd rather not have a more complicated checking logic for what is already a very rare error.

slicec/src/main.rs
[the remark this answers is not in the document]
InsertCreativityHere: Maybe this is a real concern?
The issue with path normalization is that it's fallible and messy.

InsertCreativityHere:
I updated my PR to compare _canonicalized_ paths, eliminating small differences like `.\Test.cs` vs `./Test.cs`.
This also resolves symlinks, and redundant things like `hello/../` vs `./`.

I think this is more correct, but does also add more complexity. So, maybe this is not worth doing after all...

slicec/src/main.rs
```diff
@@ -163,6 +164,30 @@ fn write_generated_file(
...
 
+/// Checks if the path of the provided `generated_file` has already been written to during this compilation.
+/// If it has, this returns an `Err` with an error describing the overwriting. Otherwise this returns `Ok`.
+/// Note that this function normalizes/canonicalizes the path before checking.
+fn check_if_file_is_overwritten<'a>(
+    generated_file: &definition_types::GeneratedFile,
+    generator: &'a Plugin,
+    written_to_paths: &mut HashMap<PathBuf, &'a Plugin>,
+) -> Result<(), slicec::diagnostics::Error> {
+    // Attempt to canonicalize the path. If that fails for any reason, fallback to using the path string as-is.
```
pepone, line 175: a corner case we don't deal with here is case insensitive filesystems.

slicec/src/main.rs
pepone: out of curiosity what is the difficulty with not overriding the file?

pepone (approved):

externl (approved):

slicec/src/main.rs
InsertCreativityHere: Yes, I didn't actually think about that, but you're right.
I'm inclined to ignore this small possibility, mostly just because there doesn't seem to be any good API for detecting this in Rust, apart from literally opening 2 handles to the file and checking if they point to the same.

Is it worth an opening an issue just to have?

slicec/src/main.rs
InsertCreativityHere: It's not that it's difficult (implementation wise), just that it would add a 2nd code path (path1: write, path2: skip), and having only one path (always write) is simpler to reason about.

To be fair, it's a very minor thing either way we do it :)

pepone: probably not, let's ignore it

PR #784, opened 2026-05-29
Report Which Plugin a Diagnostic came from

InsertCreativityHere:
This PR adds a field to our Diagnostic types tracking which plugin (if any) a diagnostic came from.
In the case of multiple plugins reporting the same error, we still only emit a single error, but it's now more clear that it was reported by multiple plugins.

The main advantage to doing this is that errors about the plugin itself (like providing a bad option) will clearly identify which plugin had the problem.

----

Here is sample output from the console:
image
And for JSON output, we now include a new `"reported_by":[...]` field.
The Task will need to be updated to properly handle this field in a follow-up PR.

### DIFF
--- a/slicec/src/compilation_state.rs
+++ b/slicec/src/compilation_state.rs
@@ -50,10 +50,12 @@ impl CompilationState {
     /// which may run identical validation. If multiple diagnostics are _exact_ duplicates, only the first will be
     /// present in the returned [`Vec`].
     pub fn get_annotated_diagnostics(&self, options: &SliceOptions) -> Vec<AnnotatedDiagnostic> {
-        let mut annotated_diagnostics = Vec::new();
+        let mut annotated_diagnostics = Vec::<AnnotatedDiagnostic>::new();
         for diagnostic in &*self.diagnostics {
             let converted = crate::diagnostics::convert_diagnostic(diagnostic, options, &self.ast, &self.files);
-            if !annotated_diagnostics.contains(&converted) {
+            if let Some(original) = annotated_diagnostics.iter_mut().find(|d| **d == converted) {
+                original.reported_by.extend(converted.reported_by);
+            } else {
                 annotated_diagnostics.push(converted);
             }
         }
--- a/slicec/src/diagnostic_emitter.rs
+++ b/slicec/src/diagnostic_emitter.rs
@@ -76,6 +76,9 @@ impl<'a> DiagnosticEmitter<'a> {
             // Emit the message with the prefix.
             writeln!(self.output, "{prefix}: {}", console::style(&diagnostic.message).bold())?;
 
+            // Emit what the diagnostic was reported by.
+            writeln!(self.output, "    Reported by: {:?}", diagnostic.reported_by)?;
+
             // If the diagnostic contains a snippet of the offending code, display it.
             if let Some(snippet) = &diagnostic.snippet {
                 self.emit_snippet(snippet)?;
@@ -109,12 +112,13 @@ impl<'a> DiagnosticEmitter<'a> {
             };
 
             let mut serializer = serde_json::Serializer::new(&mut *self.output);
-            let mut state = serializer.serialize_struct("Diagnostic", 5)?;
+            let mut state = serializer.serialize_struct("Diagnostic", 6)?;
             state.serialize_field("message", &diagnostic.message)?;
             state.serialize_field("severity", severity)?;
             state.serialize_field("snippet", &diagnostic.snippet)?;
             state.serialize_field("notes", &diagnostic.notes)?;
             state.serialize_field("error_code", &diagnostic.code)?;
+            state.serialize_field("reported_by", &diagnostic.reported_by)?;
             state.end()?;
             writeln!(self.output)?; // Separate each diagnostic by a newline character.
         }
--- a/slicec/src/diagnostic_emitter.rs
+++ b/slicec/src/diagnostic_emitter.rs
@@ -77,7 +77,7 @@ impl<'a> DiagnosticEmitter<'a> {
             writeln!(self.output, "{prefix}: {}", console::style(&diagnostic.message).bold())?;
 
             // Emit what the diagnostic was reported by.
-            writeln!(self.output, "  Reported by: {:?}", diagnostic.reported_by)?;
+            writeln!(self.output, "  Reported by: [{}]", diagnostic.reported_by.join(", "))?;
 
             // If the diagnostic contains a snippet of the offending code, display it.
             if let Some(snippet) = &diagnostic.snippet {
--- a/slicec/src/diagnostic_emitter.rs
+++ b/slicec/src/diagnostic_emitter.rs
@@ -77,7 +77,7 @@ impl<'a> DiagnosticEmitter<'a> {
             writeln!(self.output, "{prefix}: {}", console::style(&diagnostic.message).bold())?;
 
             // Emit what the diagnostic was reported by.
-            writeln!(self.output, "    Reported by: {:?}", diagnostic.reported_by)?;
+            writeln!(self.output, "  Reported by: {:?}", diagnostic.reported_by)?;
 
             // If the diagnostic contains a snippet of the offending code, display it.
             if let Some(snippet) = &diagnostic.snippet {
--- a/slicec/src/diagnostics/annotated_diagnostic.rs
+++ b/slicec/src/diagnostics/annotated_diagnostic.rs
@@ -58,7 +58,7 @@ pub fn convert_diagnostic(
     // If the diagnostic was reported by a plugin, we just use the filename of the plugin (not its entire path).
     let reported_by = diagnostic.plugin.as_deref().map_or("slicec", |plugin_path| {
         let path = std::path::Path::new(plugin_path);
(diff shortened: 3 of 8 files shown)


slicec/src/diagnostics/annotated_diagnostic.rs
```diff
@@ -9,13 +9,26 @@ use serde::Serialize;
 
 /// An annotated version of a [`Diagnostic`], whose [`DiagnosticLevel`] has been computed (taking into account any
 /// 'allow' attributes or command-line flags), and that has pre-extracted text snippets to display alongside messages.
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq)]
```
InsertCreativityHere, line 12: We implement `PartialEq` by hand now, so that we exclude the `reported_by` from the equality check.

slicec/src/diagnostics/annotated_diagnostic.rs
```diff
@@ -42,12 +55,19 @@ pub fn convert_diagnostic(
         snippet: get_snippet(&n.span, files),
     });
 
+    // If the diagnostic was reported by a plugin, we just use the filename of the plugin (not its entire path).
+    let reported_by = diagnostic.plugin.as_deref().map_or("slicec", |plugin_path| {
```
InsertCreativityHere, line 62: I went back and forth on this, but I decided that in the case where the diagnostic is issued by `slicec`, it's worthwhile to still explicitly say that this came from `slicec`, and not a plugin.

It also makes the implementation cleaner and keep the error format 100% consistent, instead of being a little different for plugin diagnostics vs those about bad syntax for example.

InsertCreativityHere: Note that `AnnotatedDiagnostic` is the "finished" form of a diagnostic.
These are always de-duplicated, and are what we're actually going to print to the user.

The normal `Diagnostic` type is what gets reported throughout the code, and before we exit, we convert the `Diagnostic`s to `AnnotatedDiagnostic`s so that we can print them.

----

Therefore `Diagnostic` can have at most one reporter (since only one thing reported it).
But `AnnotatedDiagnostic` can have multiple (since duplicate diagnostics will of been merged together).

slicec/src/compilation_state.rs
```diff
@@ -50,10 +50,12 @@ impl CompilationState {
...
     pub fn get_annotated_diagnostics(&self, options: &SliceOptions) -> Vec<AnnotatedDiagnostic> {
-        let mut annotated_diagnostics = Vec::new();
+        let mut annotated_diagnostics = Vec::<AnnotatedDiagnostic>::new();
         for diagnostic in &*self.diagnostics {
             let converted = crate::diagnostics::convert_diagnostic(diagnostic, options, &self.ast, &self.files);
-            if !annotated_diagnostics.contains(&converted) {
+            if let Some(original) = annotated_diagnostics.iter_mut().find(|d| **d == converted) {
+                original.reported_by.extend(converted.reported_by);
+            } else {
                 annotated_diagnostics.push(converted);
```
InsertCreativityHere, line 59: Previously, we completely threw away duplicate diagnostics.
Now if we encounter a duplicate, we update the original's `reported_by` before throwing away the duplicate now.

slicec/src/compilation_state.rs
```diff
@@ -50,10 +50,12 @@ impl CompilationState {
     /// which may run identical validation. If multiple diagnostics are _exact_ duplicates, only the first will be
     /// present in the returned [`Vec`].
     pub fn get_annotated_diagnostics(&self, options: &SliceOptions) -> Vec<AnnotatedDiagnostic> {
-        let mut annotated_diagnostics = Vec::new();
+        let mut annotated_diagnostics = Vec::<AnnotatedDiagnostic>::new();
```
InsertCreativityHere, line 53: I have no idea why, but the compiler insists on putting a type-hint here now.
It seems unnecessary to me, but who am I to argue with the compiler...

slicec/src/diagnostic_emitter.rs
```diff
@@ -76,6 +76,9 @@ impl<'a> DiagnosticEmitter<'a> {
             // Emit the message with the prefix.
             writeln!(self.output, "{prefix}: {}", console::style(&diagnostic.message).bold())?;
 
+            // Emit what the diagnostic was reported by.
+            writeln!(self.output, "  Reported by: [{}]", diagnostic.reported_by.join(", "))?;
```
InsertCreativityHere, line 80: See the `diagnostic_output_test` or my PR description for an example of what the output looks like.

slicec/src/main.rs
```diff
@@ -274,8 +274,11 @@ fn main() -> ExitCode {
                 }
             }
 
-            // Store generator's diagnostics for later emission.
-            diagnostics.extend(generator_diagnostics.into_inner());
+            // Store generator's diagnostics for later emission, after setting the plugin that reported them.
+            for mut generator_diagnostic in generator_diagnostics.into_inner() {
+                generator_diagnostic.plugin = Some(generator.path.clone());
```
InsertCreativityHere, line 279: Where we actually set `plugin`.
This is after we've finished (or tried) to run the plugin and are storing its diagnostics.

slicec/src/slice_file_converter.rs
```diff
@@ -69,7 +69,7 @@ pub fn convert_diagnostic(diagnostic: Diagnostic, ast: &Ast, files: &[GrammarSli
     let (span, scope) = convert_source(diagnostic.source.as_deref(), ast, files, output);
 
     // Construct and push the converted 'slicec' `Diagnostic`.
-    let converted_diagnostic = slicec::diagnostics::Diagnostic { kind, span, scope, notes };
+    let converted_diagnostic = slicec::diagnostics::Diagnostic { kind, span, scope, notes, plugin: None };
```
InsertCreativityHere, line 72: We construct the diagnostic in one-shot here, so had to update it to include the new field.
We manually set it to it's default value (`None`).

Rust doesn't have the idea of default field values like other languages.

slicec/tests/diagnostic_output_tests.rs
```diff
@@ -156,9 +180,13 @@ error [E008]: invalid enum 'E': enums must contain at least one enumerator
         // Assert
         assert_eq!(converted_diagnostics.len(), 4);
         assert_eq!(converted_diagnostics[0].message, "This is a test");
```
InsertCreativityHere, line 178: Check that the `reported_by`s are correct after merging.

slicec/src/compilation_state.rs
```diff
@@ -50,10 +50,12 @@ impl CompilationState {
...
         for diagnostic in &*self.diagnostics {
             let converted = crate::diagnostics::convert_diagnostic(diagnostic, options, &self.ast, &self.files);
-            if !annotated_diagnostics.contains(&converted) {
+            if let Some(original) = annotated_diagnostics.iter_mut().find(|d| **d == converted) {
+                original.reported_by.extend(converted.reported_by);
```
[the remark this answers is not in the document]
InsertCreativityHere: Yeah that's fine. A single plugin reporting duplicate diagnostics is a flaw in the plugin itself.

In this case, silently deduplicating is actually wrong IMO. Whereas this new behavior would actually make it clear that "hey, there's something wrong with this plug" (if we ever see `Reported by: ["foo", "foo"]`).

slicec/src/diagnostics/annotated_diagnostic.rs
```diff
@@ -42,12 +55,19 @@ pub fn convert_diagnostic(
...
+        let path = std::path::Path::new(plugin_path);
+        path.file_stem().unwrap().to_str().unwrap()
+    });
```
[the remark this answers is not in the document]
InsertCreativityHere: So technically you're right, and I'll see about avoiding the unwrap (if it doesn't add too much clutter).

But, I don't think that correctly handling "I passed in `.` as my plugin" is a real concern here...

slicec/src/diagnostics/diagnostic.rs
```diff
@@ -18,6 +18,9 @@ pub struct Diagnostic {
 
     /// Any additional information that should be reported alongside this diagnostic's main message.
     pub notes: Vec<Note>,
+
+    /// The plugin that reported this diagnostic, or `None` if this diagnostic was reported by 'slicec' itself.
```
bernardnormier, line 22: We use the term `plugin` and not `generator` in slicec?

bernardnormier (approved):

externl (approved):

slicec/src/diagnostics/diagnostic.rs
InsertCreativityHere: We use both terms in `slicec`.
A `plugin` is any binary that logically implements a Slice operation, and that is made to be forked by `slicec`.
A `generator` is a specific kind of plugin; a plugin that implements `generateCode`.

Right now `generators` are the only kind of `plugin` we have. But we wanted to keep this design flexible for future kinds of plugins, back when we were discussing it. I've been careful to refer to `generator`s for parts of the code specific to code-gen, but here, "reporting diagnostics" isn't specific to code-gen. Any plugin could do that.

PR #788, opened 2026-06-01
Updated the LICENSEs, READMEs, and Other Small Things

InsertCreativityHere:
Probably best to review the Markdown Previews in VSCode for this PR...

This PR primarily:
- improves the README for the entire repo (which is visible on the GitHub landing page)
- updates the README for `slicec` (which is baked into the crate and visible on its `crates.io` page)
- adds a README for `slice-codec` ( --- same as above --- )

----

It also adds a LICENSE to `slice-codec` and changes the LICENSE in `slicec` to be a copy instead of a symlink.
Symlinks in Git seem to be moere trouble that they're worth, which is surprising since most multi-crate repos use symlinked LICENSE files \*shrugs\*.


.github/workflows/ci.yml
```diff
@@ -41,7 +41,6 @@ jobs:
           "",
           "alloc",
           "std",
-          "tokio",
```
InsertCreativityHere, line 44: While documenting the crate features for `slice-codec`, I realized we still had all the cruff about `tokio` sitting around. This should of been removed before, but slipped past us.

Maybe one day it will change, but right now `slice-codec` has no special handling for `tokio`, and this feature is 100% dead code.

slice-codec/LICENSE
```diff
@@ -0,0 +1 @@
+../LICENSE
```
InsertCreativityHere: `crates.io` understands this syntax. This way we only need the full LICENSE file in one place, and each crate knows to yank a copy of it when being published.

slicec/README.md
```diff
@@ -4,41 +4,119 @@
 
 # The Slice compiler (slicec)
```
InsertCreativityHere, line 6: Don't bother looking at the old README very hard.
It was completely rewritten from scratch.

slice-codec/LICENSE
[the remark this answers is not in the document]
InsertCreativityHere: Maybe it would be best to change this to a proper symlink.
Although to my knowledge, `cargo` does handle this correctly. But for someone looking at the source code, obviously this will be less useful...

slice-codec/README.md
```diff
@@ -0,0 +1,98 @@
...
+fn main() -> slice_codec::Result<()> {
+    let mut buffer: Vec<u8> = Vec::new();
+
+    // Encode some values into a byte buffer.
+    let mut encoder = Encoder::from(&mut buffer);
+    encoder.encode("hello")?;
+    encoder.encode::<i32>(42)?;
+
+    // Decode them back out.
+    let mut decoder = Decoder::from(&buffer);
```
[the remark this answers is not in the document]
InsertCreativityHere: I made sure to run this as a doc-test, and the code compiled and ran just fine.
Changing the `42` to `4` in the assert caused it to fail, so it's not like it was skipped over.
CoPilot is just wrong here.

slice-codec/LICENSE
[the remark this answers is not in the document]
InsertCreativityHere: I changed to use symlinks.
Just feels cleaner.

slice-codec/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Changelog
```
bernardnormier: See my comment on the Changelog PR.

slice-codec/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
bernardnormier: We should definitely have a single top-level LICENSE file and no other license in this repo.

slicec/README.md
```diff
@@ -4,41 +4,119 @@
...
-and converting them into a typed AST. It also exposes APIs for traversing and searching this AST, as well as retrieving
-any warnings / errors discovered in the parsed Slice definitions.
+The `slicec` binary is a command-line tool that takes Slice files, parses them into a typed _Abstract Syntax Tree_
+(AST), validates them, and then encodes them using the [Slice encoding][slice-encoding] so other tools can consume it.
 
-Crates can include the `slicec` library by listing it as a dependency:
+Code generation is performed by **generator plugins**: external executables that are passed with `--generator` (or `-G`)
+on the command line. `slicec` spawns each generator as a subprocess, sends the Slice-encoded AST over its `stdin`, and
+waits for a response over its `stdout`. This response consists of generated files (which `slicec` will then write to the
+appropriate location) or diagnostics (which are collected and reported by `slicec` at the end).
```
bernardnormier: _and_ diagnostics is more accurate.

slicec/README.md
```diff
@@ -4,41 +4,119 @@
...
+
+# Run the tests.
+cargo test -p slicec
 ```
 
-[ci-home]: https://github.com/icerpc/slicec/actions/workflows/ci.yml
-[license]: https://github.com/icerpc/slicec/slicec/blob/main/LICENSE
+## License
+
+Licensed under the [Apache License, Version 2.0][license].
```
bernardnormier: Should be:
```
slicec is licensed under the [Apache License version 2.0][...].
```

slice-codec/README.md
```diff
@@ -0,0 +1,98 @@
...
+# Build with the default features.
+cargo build -p slice-codec
+
+# Run the tests with all features enabled.
+cargo test -p slice-codec --all-features
+```
+
+## License
+
+Licensed under the [Apache License, Version 2.0][license].
```
bernardnormier: See other README.

slice-codec/README.md
```diff
@@ -0,0 +1,98 @@
...
+slice-codec = { version = "0.4", default-features = false }
+```
+
+If you have a global allocator and need to decode owned types like `String` or `Vec`, also enable the `alloc` feature.
+
+### Feature Flags
+
+| Feature  | Default | Description                                                                                                                      |
+| -------- | :-----: | ---------------------------------------------------------------------------------------------------------------------------------|
+| `std`    |   ✅    | Enables implementations to encode/decode standard-library types such as `String`, `Vec`, and `HashMap`. Implies `alloc`.         |
```
bernardnormier: Does not appear to be correctly aligned.

slice-codec/LICENSE
bernardnormier: Please don't use symbolic links.

InsertCreativityHere: Yes, I reverted my symbolic links after deciding it was just a headache.
We only have a single full `LICENSE` at the root of the repo.

Each subfolder (each crate that is) has a `LICENSE` which just points cargo at the top-level one.
It's probable that we can even remove these and just use configuration to point at the right LICENSE.
Will look more in the morning.

slice-codec/README.md
InsertCreativityHere: Fixed.
Even in monospace font, the checkmark emoji doesn't take up the same amount of space, which threw me off.

slicec/README.md
InsertCreativityHere: Changed as suggested.

slicec/README.md
InsertCreativityHere: Fixed.

slice-codec/README.md
InsertCreativityHere: And fixed alike.

bernardnormier (approved):
The updated READMEs look fine. Please delete the new CHANGELOG file.

pepone (approved):

externl (approved):

PR #790, opened 2026-06-02
Updated References to 'main' Branch

InsertCreativityHere:


.github/workflows/coverage.yml
```diff
@@ -2,7 +2,7 @@ name: coverage
 
 on:
   push:
-    branches: ["main"]
```
InsertCreativityHere, line 5: This used to be in quotes... I removed the quotes since the other CI files didn't use quotes.
There's no way this breaks anything right?

InsertCreativityHere: Ignore this. Apparently single quotes are the way to go now...

externl (approved):

PR #791, opened 2026-06-02
Fix Quotes in Workflow Files (Single is Preferred)

InsertCreativityHere:
Inspired by #790.


externl (approved):
Looks fine. Generally you don't use quotes for strings like main. But it doesn't really matter.

PR #792, opened 2026-07-10
House-keeping Improvements

InsertCreativityHere:
While giving a fresh slate to https://github.com/icerpc/icerpc-rust/pull/26, I decided to look back into some of the things we were setting/doing in these ancillary files, and sure enough there was some cleanup/improvement to do.

They're all minor enough that I think not backporting anything is perfectly fine.


.github/workflows/ci.yml
```diff
@@ -10,7 +10,8 @@ permissions:
   contents: read
 
 env:
-  WARNINGS_AS_ERRORS: yes
```
InsertCreativityHere, line 13: This was apparently useless?
Maybe we had some custom logic using it in the past, but nothing in Rust checks or cares about this environment variable.

.github/workflows/ci.yml
```diff
@@ -30,8 +31,6 @@ jobs:
         run: cargo clippy --all-targets --no-deps -p slicec
       - name: 📚 Generate Documentation
         run: cargo doc --document-private-items --no-deps -p slicec
-        env:
-          RUSTDOCFLAGS: "-D warnings"
```
InsertCreativityHere, line 34: Better to set this for all jobs, instead of doing it for every single one.

.github/workflows/spellcheck.yml
```diff
@@ -15,4 +17,4 @@ jobs:
       - uses: streetsidesoftware/cspell-action@v8
         with:
           files: |
-            **/*.{rs,md,json}
+            **/*.{rs,slice,md,json,yml,toml}
```
InsertCreativityHere, line 20: Spell check everything!

.github/workflows/spellcheck.yml
```diff
@@ -1,8 +1,10 @@
 name: "Check spelling"
 
 on:
-  pull_request:
   push:
+    branches: ['main']
+  pull_request:
+    branches: ['main']
```
InsertCreativityHere, line 8: For parity with our other workflow files.

.vscode/cspell.json
InsertCreativityHere, line 1: I went through and checked which of these words was actually still necessary, since we've pruned and removed alot of code from slicec after dropping Slice1 support.

Sure enough, many of them are no longer necessary.

.gitignore
```diff
@@ -2,10 +2,6 @@
 # will have compiled files and executables
 target/
 
-# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
-# More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
-Cargo.lock
```
InsertCreativityHere, line 7: The Rust team has softened their stance about including the Cargo.lock file.
Including it gives us more reproducible builds but does some harm to the ecosystem since we won't be testing against the latest and greatest anymore.

By making sure to update our dependencies more often, this becomes a non-issue though.

slice-codec/tests/encoding_tests.rs
```diff
@@ -76,7 +76,7 @@ mod fixed_size {
         );
         let remaining = &buffer[expected.len()..];
         assert!(
```
InsertCreativityHere, line 78: Also fixed some random issues I found by running the latest version of the linter.
All very minor things.

.vscode/cspell.json
```diff
@@ -12,27 +12,19 @@
...
-    "nonterminal",
     "peekable",
     "repr",
     "rfind",
-    "rsplit",
     "RUSTDOCFLAGS",
+    "RUSTFLAGS",
     "rustfmt",
-    "rustup",
     "slicec",
```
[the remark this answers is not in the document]
InsertCreativityHere: Screenshot 2026-07-10 at 21 32 05

To be fair I was also thinking this would happen : v)
But, I guess `rustup` is in the Rust dictionary we pull in, which would be pretty reasonable.

Issue #798, opened 2026-08-19
`enum` with an Anonymous Underlying Type Causes a Panic


InsertCreativityHere:
`grammar.rs:271: underlying_type.map(|t| t.downcast::<Primitive>().unwrap())`.
`TypeRef::downcast` returns `Err` for a patched type mismatch, and anonymous built-ins are eagerly patched, so:

```
module T
enum E : Sequence<int32> { A }
```

panics (`Result::unwrap()` on `Err`) instead of emitting "enum underlying type must be an integer primitive".
Also `Dictionary<…>`, `Result<…>`.

(A named/alias underlying is unpatched → handled gracefully.)

InsertCreativityHere:
The same bug is also present for interfaces which inherit from anonymous types:

```slice
module T
interface I : Sequence<int32> {}
```
Same pattern at `grammar.rs:142: base.downcast::<Interface>().unwrap()`.
panics instead of "interfaces can only inherit from interfaces". Fix both by matching on the Err and pushing a diagnostic.

InsertCreativityHere:
This was fixed in #802.

PR #802, opened 2026-08-20
Fix Panics from Using Anonymous Types as a Base or Underlying Type

InsertCreativityHere:
In a Slice file, if you try to use an anonymous type (a sequence, dictionary or result type) as:
- the underlying type of an enum (`enum Foo : Sequence<int8>`), or
- the base of an interface (`interface Foo : Sequence<int8>`)

It would cause the compiler to crash, instead of correctly reporting an error.
The same would also happen when using a primitive type as the base of an interface (`interface Foo : uint8`).

These were all due to a bad assumption we made about these types being unpatched during the parsing phase.
But anonymous types and primitives _are_ patched at construction (even during parsing) since they're fully knowable.

This PR fixes both parts of #798.

----

## What's Changed entry

None — this construct was already disallowed; only the rejection changed from a panic to a proper error. And this construct has probably never been written in the wild, so this affects no users.

### DIFF
--- a/slicec/src/grammar/elements/type_ref.rs
+++ b/slicec/src/grammar/elements/type_ref.rs
@@ -51,7 +51,10 @@ impl<T: Element + ?Sized> TypeRef<T> {
 impl<T: Type + ?Sized> TypeRef<T> {
     // This intentionally shadows the trait method of the same name on `Type`.
     pub fn type_string(&self) -> String {
-        let mut s = self.definition().type_string();
+        let mut s = match &self.definition {
+            TypeRefDefinition::Patched(ptr) => ptr.borrow().type_string(),
+            TypeRefDefinition::Unpatched(ident) => ident.value.clone(),
+        };
         if self.is_optional {
             s += "?";
         }
--- a/slicec/src/parsers/slice/grammar.rs
+++ b/slicec/src/parsers/slice/grammar.rs
@@ -144,7 +144,8 @@ fn construct_interface(
             if interface_ref.is_err() {
                 Diagnostic::from_error(Error::TypeMismatch {
                     expected: "interface".to_owned(),
-                    actual: base.type_string(),
+                    // 'kind' is safe to call here. `downcast` can only fail if the type is already patched.
+                    actual: base.definition().kind().to_owned(),
                     is_concrete: true,
                 })
                 .set_span(&base.span)
@@ -286,7 +287,8 @@ fn construct_enum(
         if primitive_ref.is_err() {
             Diagnostic::from_error(Error::EnumUnderlyingTypeNotSupported {
                 enum_identifier: identifier.value.clone(),
-                kind: Some(type_ref.type_string()),
+                // 'kind' is safe to call here. `downcast` can only fail if the type is already patched.
+                kind: Some(type_ref.definition().kind().to_owned()),
             })
             .set_span(&type_ref.span)
             .push_into(parser.diagnostics);
--- a/slicec/src/parsers/slice/grammar.rs
+++ b/slicec/src/parsers/slice/grammar.rs
@@ -139,8 +139,21 @@ fn construct_interface(
     let bases = bases
         .unwrap_or_default() // Create an empty vector if no bases were specified.
         .into_iter()
-        .map(|base| base.downcast::<Interface>().unwrap())
+        .filter_map(|base| {
+            let interface_ref = base.downcast::<Interface>();
+            if interface_ref.is_err() {
+                Diagnostic::from_error(Error::TypeMismatch {
+                    expected: "interface".to_owned(),
+                    actual: base.type_string(),
+                    is_concrete: true,
+                })
+                .set_span(&base.span)
+                .push_into(parser.diagnostics);
+            }
+            interface_ref.ok()
+        })
         .collect::<Vec<_>>();
+
     let comment = parse_doc_comment(parser, &identifier.value, raw_comment);
 
     let mut interface_ptr = OwnedPtr::new(Interface {
@@ -268,7 +281,19 @@ fn construct_enum(
     enumerators: Vec<OwnedPtr<Enumerator>>,
     span: Span,
 ) -> OwnedPtr<Enum> {
-    let underlying = underlying_type.map(|type_ref| type_ref.downcast::<Primitive>().unwrap());
+    let underlying = underlying_type.and_then(|type_ref| {
+        let primitive_ref = type_ref.downcast::<Primitive>();
+        if primitive_ref.is_err() {
+            Diagnostic::from_error(Error::EnumUnderlyingTypeNotSupported {
+                enum_identifier: identifier.value.clone(),
+                kind: Some(type_ref.type_string()),
+            })
+            .set_span(&type_ref.span)
+            .push_into(parser.diagnostics);
+        }
+        primitive_ref.ok()
+    });
+
     let comment = parse_doc_comment(parser, &identifier.value, raw_comment);
 
     let mut enum_ptr = OwnedPtr::new(Enum {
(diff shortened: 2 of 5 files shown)


slicec/src/grammar/elements/type_ref.rs
```diff
@@ -51,7 +51,10 @@ impl<T: Element + ?Sized> TypeRef<T> {
 impl<T: Type + ?Sized> TypeRef<T> {
     // This intentionally shadows the trait method of the same name on `Type`.
     pub fn type_string(&self) -> String {
```
InsertCreativityHere, line 53: This is a tangential change.
It lets us print a reference _before_ the patching phase has run.
i.e. now we can safely print a reference while the parser is running.

For unpatched references, we print the exact text found in the Slice file.
For patched references, the behavior is unchanged.

slicec/src/parsers/slice/grammar.rs
```diff
@@ -139,8 +139,21 @@ fn construct_interface(
     let bases = bases
         .unwrap_or_default() // Create an empty vector if no bases were specified.
         .into_iter()
-        .map(|base| base.downcast::<Interface>().unwrap())
+        .filter_map(|base| {
+            let interface_ref = base.downcast::<Interface>();
```
InsertCreativityHere, line 143: Now we check the result of `downcast::<Interface>` instead of assuming it would succeed.
If it `is_err()` we report and error and discard the `base`. Otherwise it gets kept like before.

slicec/src/parsers/slice/grammar.rs
```diff
@@ -268,7 +281,19 @@ fn construct_enum(
     enumerators: Vec<OwnedPtr<Enumerator>>,
     span: Span,
 ) -> OwnedPtr<Enum> {
-    let underlying = underlying_type.map(|type_ref| type_ref.downcast::<Primitive>().unwrap());
+    let underlying = underlying_type.and_then(|type_ref| {
+        let primitive_ref = type_ref.downcast::<Primitive>();
```
InsertCreativityHere, line 286: Same for enums, we check if the `downcast` succeeded instead of blindly calling `unwrap` (assuming it will always work).

slicec/src/parsers/slice/grammar.rs
```diff
@@ -268,7 +281,19 @@ fn construct_enum(
...
+        if primitive_ref.is_err() {
+            Diagnostic::from_error(Error::EnumUnderlyingTypeNotSupported {
+                enum_identifier: identifier.value.clone(),
+                kind: Some(type_ref.type_string()),
+            })
+            .set_span(&type_ref.span)
+            .push_into(parser.diagnostics);
+        }
+        primitive_ref.ok()
+    });
```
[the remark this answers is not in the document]
InsertCreativityHere: I considered doing this, but didn't, since it won't make a difference.

If _any_ error is reported during parsing time (including the two added in this PR), the compiler exits early,
so we will never even reach the validation logic you're worried about falsely triggering.

bernardnormier:
The code looks good — a few nits on the PR description:

1. **Auto-close syntax:** "This PR fixes both parts of #798." won't auto-close the issue on merge; GitHub needs the exact keyword form. Please use `Fixes #798` on its own line — if "both parts" matters, "Fixes #798 (both the enum and interface cases)."

2. **Tone:** please rewrite point A in more neutral language, e.g. *"This construct doesn't appear in any realistic Slice definition."* PR descriptions are part of the permanent public record.

3. **Broken sentence:** "In a Slice file, if you try to use … as: [bullets] It would cause…" fractures mid-sentence. Suggested: *"Using an anonymous type (sequence, dictionary, or result) as the underlying type of an enum (`enum Foo : Sequence<int8>`) or the base of an interface (`interface Foo : Sequence<int8>`) caused the compiler to panic instead of reporting an error."*

4. **Name the bad assumption:** "a bad assumption we were making about these during the parsing phase" is the sentence a future reader most wants — say it concretely: `construct_interface` and `construct_enum` unwrapped a downcast that assumed bases/underlying types are never anonymous types.

5. **Title:** enums don't "inherit" an underlying type. Consider "Fix panic when an anonymous type is used as an enum's underlying type or an interface's base."

6. **What's Changed entry:** please use the section format from the icerpc-csharp convention instead of the freeform "This PR doesn't have any notes, since: A)… B)…":

   ```markdown
   ## What's Changed entry

   None — this construct was already disallowed; only the rejection changed from a panic to a proper error.
   ```

   (Agreed that `None` is the right call here.)

externl (commented):
The two `unwrap()` sites are the right ones to fix, but reporting from the parser has some knock-on effects worth considering.

- `CompilationState::apply` skips both `patch_ast` and `validate_ast` once any error exists, so a parse-time diagnostic here suppresses every patcher and validator diagnostic in the compilation — `enum E : Sequence<Missing>` reports the underlying-type error but never that `Missing` is undefined. Noting it for the record; I don't think it's important or worth fixing, since you've already hit an error either way.
- `type_string()` appends `?` for optional refs, so `enum E : Dictionary<string, int8>?` reports `underlying type 'Dictionary<string, int8>?' is not supported`. `kind` is documented as a kind name, and the validator raising the same E009 fills it with `definition().kind()`, which never carries optionality. Worth stripping the `?` here. (These enums also lose the separate `CannotUseOptionalUnderlyingType` error, but that follows from the point above rather than from anything this line does.)
- Relatedly, `TypeMismatch::actual` now holds two different things: `downcast()` returns `Ok` unconditionally for `Unpatched` refs, so named bases still reach the patcher and get kind names while primitives and anonymous types get type strings. `interface I : S` reads `expected an 'interface' but found a 'struct'`, but `interface I : Sequence<bool>` reads `found a 'Sequence<bool>'`. The new test cases encode both spellings side by side.
- Minor: E009's span is the whole enum when the validator reports it, but just the type ref here.

Also worth updating the title/body — the interface fix covers primitives too (`interface I : uint8` panicked on `main`), not just anonymous types.

InsertCreativityHere:
Response to @bernardnormier:

1) I'm going to manually close the issue after it's been fixed, instead of purely relying on GitHub's auto-close.
Which is fine to do. I think mentioning "both parts" is important here; since there's 2 sub-issues in it.
Clobbering this fact, or adding a redundant line just hurts the PR description, for negligible benefit.

2) Fair point, I reworded it. But I think "realistic" is too soft. It's not just _unrealistic_ to have an interface
inherit from a sequence; It's straight-up wrong.

3) This was originally what I wrote (more or less). But the sentence is long and hard to parse. If you read my PR,
but just remove the bullets, it's a perfectly valid and complete sentence. The bullets just help to 'break' up the long
sentence, and to clearly delineate the two separate items it's talking about.

4) Stating the specific functions is too in the weeds for a PR description, in my opinion. If someone wants to know what
code changed, they should review the actual changes.

5) I did consider a more technically correct title, but it gets too long. The title you suggest is almost double the
length, and it's good to keep GitHub titles from getting so long. I think everyone will understand what the title says,
even if it's not 100% technically accurate. And I posit that's good enough for just being a title; not a code change.

6) Updated the description to include the section as suggested, and will do so via template in the future.

InsertCreativityHere:
Response to @externl

1) Yep, this PR actually relies on the fact that a parse-time error early exits the compiler. Otherwise, downstream code
would need to handle degenerate cases like this one.

2) I did a "drive-by" fix for the E009 span you found. Now it correctly only points to the underlying type, instead of
the whole enum. Not really related to this PR, but still good to fix!

3) I switched to use `kind` instead of `type_string` like you said. Originally I didn't because it wasn't clear it was
safe to do here, since nothing is patched. But since this only affects anonymous types, which don't require patching,
it is indeed safe. Just not obviously so... So I added a comment too!

bernardnormier (approved):

externl (approved):
Looks good!

PR #813, opened 2026-08-31
Fix Ambiguity in Elements using Escaped Primitive Keywords

### LINKED ISSUE #808: Improper Handling of Keyword Escaped Identifiers
Slice allows you to use keywords as an identifier provided you escape them:
```
module \int32
```
However, the way that we handle them under the hood is incorrect and buggy.
This is because right now the `\` backslash only exists at lexing time: when the lexer sees an identifier, normally it first performs a lookup against a list of keywords, and only if there is no match do we skip. Either way, the lexer eats the `\`, and from then on `int32` is indistinguishable from the keyword name.

----

# Proposed fix:

We should keep the backslash around as part of the identifier, so that `\int32` and `int32` can be distinguished not just by the lexer, but by the redefinition checker, the type-patcher, etc.

For correctness' sake though, we should still eat the `\` in cases where it's non-useful. For example, it's valid to escape something that isn't a keyword (with the rationale being consistency and because keywords may be added/removed in the future). But, it's important to remove `\` in these cases because:
```
"\Foo" and "Foo" => there should be no difference between these identifiers
"\bool" and "bool" => there should be a difference between these identifiers
```

### LINKED ISSUE #815: Doc-Comments can Reference Primitive Types as "Globally Scoped"
Right now, the Slice compiler will allow this syntax:
```
/// Here is {@link ::int32}!
```
as being equivalent to:
```
/// Here is {@link int32}!
```

This is wrongheaded though, writing `::int32` should not ever be treated as a primitive type, since primitive types do not have scopes. A leading `::` implies a user-defined type, and the doc-comment parser should be enforcing this.

InsertCreativityHere:
This PR fixes #808, which made it possible to 'overwrite' primitive types

```
module \int32
typealias Foo = int32 // This references the module now, not the primitive type!
```

It also fixes #815, which let users reference a primitive type as `::int32` in doc-comments, instead of treating that as a globally-scoped user-defined element.

----

But it does not use the proposed solution. It turns out there is a much simpler solution thanks to the fact:
"the Slice Lexer immediately patches primitives types". `TypeRef` can be in 2 states, either `Patched` or `Unpatched`. For user-defined types, the parser creates them `Unpatched`, and then after we've parsed everything, we lookup the names and patch all the types. But for primitives, we patch them immediately, since we already know what they are.

This means: We never need to lookup a primitive type by name. So, this PR drops the primitives from the lookup table entirely.
Making it absolutely impossible for a conflict to ever happen. Instead, we now have a dedicated function for looking up the primitive types, which bypasses all the lookup machinery.

## What's Changed

This PR changes the following behaviors:

- `\xxx` can no longer be used to reference the primitive type `xxx`. Now placing a backslash in front of a keyword means the escaped identifier will _always_ be resolved against user-defined types and never treated as a primitive.
- Fixed a bug where defining a module whose identifier was an escaped primitive type, would cause future resolutions of that primitive type to fail. For example:
```
module \int32
struct Foo { f: int32 } // `f` would be referencing the module, not the primitive here!
```
- Fixed a bug where primitive types could be referenced with a leading scope in doc-comments.

### DIFF
--- a/slicec/src/ast/mod.rs
+++ b/slicec/src/ast/mod.rs
@@ -13,7 +13,7 @@ use std::collections::HashMap;
 /// slice files passed into the compiler.
 ///
 /// The AST is primarily for centralizing ownership of Slice elements, but also features lookup functions for finding
-/// nodes (see [`find_node_with_scope`](Ast::find_node_with_scope)) and their
+/// nodes (see [`find_node_by_id`](Ast::find_node_by_id)) and their
 /// elements (see [`find_symbol_by_id`](Ast::find_symbol_by_id)).
 ///
 /// In practice, there is a single instance of the AST per compilation, which is [created](Ast::create) during
@@ -118,7 +118,7 @@ impl Ast {
     ///
     /// If a node can be found with the provided identifier, this returns a reference to its [node](Node) in the AST
     /// wrapped in `Ok`. Otherwise, this returns `Err` with a string describing why the lookup failed.
-    pub fn find_node_with_scope<'a>(&'a self, identifier: &str, scope: &str) -> Result<&'a Node, LookupError> {
+    pub fn find_node_by_id<'a>(&'a self, identifier: &str, scope: &str) -> Result<&'a Node, LookupError> {
         // If the identifier isn't globally scoped, we check for it in the provided scope,
         // followed by each of its parent scopes, until finally landing at global scope.
         if !identifier.starts_with("::") {
--- a/slicec/src/ast/mod.rs
+++ b/slicec/src/ast/mod.rs
@@ -103,11 +103,11 @@ impl Ast {
     /// working outwards through each of its parent scopes until reaching global scope.
     ///
     /// This returns the first matching AST node it can find. If another node in a more outward scope also has the
-    /// specified identifier, it is shadowed, and will not be returned.
-    ///
-    /// Anonymous types (those without identifiers) cannot be looked up. These are results, sequences, and dictionaries.
-    /// Primitive types can be looked up by their Slice keywords. Care should be taken when looking up modules (which
+    /// specified identifier, it is shadowed, and will not be returned. Exercise care when looking up modules (which
     /// can be re-opened) or parameters and return members (which share an AST scope), since these may not be unique.
+    ///
+    /// Primitive types (`int32`, `string`, etc.) and anonymous types (results, sequences, and dictionaries)
+    /// cannot be looked up with this method.
     ///
     /// This is a low level method used for retrieving nodes from the AST directly.
     /// Only use this if you need access to the node, or the pointer, holding a slice element.
@@ -129,8 +129,8 @@ impl Ast {
             // If it doesn't exist, keep checking for it in parent scopes until all enclosing scopes have been checked.
             while !scopes.is_empty() {
                 let candidate = scopes.join("::") + "::" + identifier;
-                if let Some(i) = self.lookup_table.get(&candidate) {
-                    return Ok(&self.elements[*i]);
+                if let Ok(node) = self.lookup_node_by_id(&candidate) {
+                    return Ok(node);
                 }
                 // Pop the last scope segment off to get to the next highest scope.
                 scopes.pop();
@@ -142,11 +142,7 @@ impl Ast {
         // Remove any leading '::' from the identifier, since the lookup table doesn't store them.
         // TODO switch to 'trim_prefix' (https://github.com/rust-lang/rust/issues/142312) when it's stabilized.
         let stripped_identifier = identifier.strip_prefix("::").unwrap_or(identifier);
-        let Some(index) = self.lookup_table.get(stripped_identifier) else {
-            let identifier = stripped_identifier.to_owned();
-            return Err(LookupError::DoesNotExist { identifier });
-        };
-        Ok(&self.elements[*index])
+        self.lookup_node_by_id(stripped_identifier)
     }
 
     /// Returns a reference to a Slice symbol (user-defined element) with the provided identifier and specified type,
@@ -163,11 +159,7 @@ impl Ast {
     where
         &'a T: TryFrom<&'a Node, Error = LookupError>,
     {
-        let Some(index) = self.lookup_table.get(identifier) else {
-            let identifier = identifier.to_owned();
-            return Err(LookupError::DoesNotExist { identifier });
-        };
-        (&self.elements[*index]).try_into()
+        self.lookup_node_by_id(identifier)?.try_into()
     }
 
     /// Returns an immutable slice of all the [nodes](Node) contained in this AST.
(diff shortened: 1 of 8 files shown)


slicec/src/grammar/elements/primitive.rs
```diff
@@ -92,3 +94,29 @@ impl Element for Primitive {
         }
     }
 }
+
+impl FromStr for Primitive {
```
InsertCreativityHere, line 97: Lets you convert a `string` into a `Primitive`. Added for convenience.

slicec/src/parsers/slice/grammar.rs
```diff
@@ -415,10 +415,9 @@ fn construct_type_ref(
 }
 
 fn primitive_to_type_ref_definition(parser: &Parser, primitive: Primitive) -> TypeRefDefinition {
```
InsertCreativityHere, line 417: The one place we were looking up primitive types.
Now it uses the dedicated `find_primitive_node`, instead of the old `find_node`.

slicec/src/patchers/comment_link_patcher.rs
```diff
@@ -163,8 +167,7 @@ fn convert_node_to_entity_ptr(node: &Node) -> Result<WeakPtr<dyn Entity>, String
 
         Node::Module(_) => Err("modules cannot be linked to".to_owned()),
         Node::Parameter(_) => Err("parameters cannot be linked to".to_owned()), // TODO improve for return members.
-        Node::Primitive(_) => Err("primitive types cannot be linked to".to_owned()),
```
InsertCreativityHere, line 166: These are matching against the result of `find_node`, which now can never return a `Primitive`. So this branch is dead code.

slicec/src/patchers/comment_link_patcher.rs
```diff
@@ -101,13 +102,16 @@ impl CommentLinkPatcher<'_> {
...
         };
-
         // Look up the linked-to entity in the AST.
         let result = ast
-            .find_node_with_scope(&identifier.value, &commentable.parser_scoped_identifier())
-            .map_err(|lookup_error| match lookup_error {
-                LookupError::DoesNotExist { identifier } => format!("no element named '{identifier}' exists in scope"),
-                _ => unreachable!("`find_node_with_scope` reported an error other than `DoesNotExist`"),
+            .find_node_by_id(&identifier.value, &commentable.parser_scoped_identifier())
+            .map_err(|lookup_error| {
```
InsertCreativityHere, line 108: This lookup will now fail for either: undefined types, or primitive types.
Good to have dedicated error messages for both cases.

slicec/src/ast/mod.rs
```diff
@@ -13,8 +13,8 @@ use std::collections::HashMap;
 /// slice files passed into the compiler.
 ///
 /// The AST is primarily for centralizing ownership of Slice elements, but also features lookup functions for finding
-/// nodes (see [`find_node`](Ast::find_node) and [`find_node_with_scope`](Ast::find_node_with_scope)) and their
-/// elements (see [`find_symbol_by_id`](Ast::find_symbol_by_id)).
+/// nodes (see [`find_node_by_id`](Ast::find_node_by_id)) and their elements
```
InsertCreativityHere, line 16: We used to have 2 functions: `find_node` and `find_node_with_scope`.

`find_node` was only called to lookup primitives, so it was renamed `find_primitive_node`.
`find_node_with_scope` was renamed to `find_node_by_id` to be symmetric with `find_symbol_by_id`.

slicec/src/ast/mod.rs
```diff
@@ -70,44 +69,14 @@ impl Ast {
             Node::Primitive(OwnedPtr::new(Primitive::String)),
         ];
 
-        let lookup_table = HashMap::from([
```
InsertCreativityHere, line 73: We no longer seed the primitive types in the lookup table, since they are never looked up anymore.
This is the core thing that fixes the problem where `module \int32` would overwrite this table's entries.

slicec/src/ast/mod.rs
```diff
@@ -117,40 +86,29 @@ impl Ast {
...
     /// ```
-    pub fn find_node<'a>(&'a self, identifier: &str) -> Result<&'a Node, LookupError> {
-        self.lookup_table
-            .get(identifier)
-            .map(|i| &self.elements[*i])
-            .ok_or_else(|| LookupError::DoesNotExist {
-                identifier: identifier.to_owned(),
-            })
+    pub fn find_primitive_node(&self, primitive: Primitive) -> &Node {
+        self.elements.get(primitive as usize).expect("Missing primitive node!")
```
InsertCreativityHere, line 93: The order of the primitives is guaranteed. We know their exact indices.

slicec/src/ast/mod.rs
```diff
@@ -271,6 +218,15 @@ impl Ast {
         // Add the element to this AST.
         self.add_element(element)
     }
+
+    fn lookup_node_by_id<'a>(&'a self, identifier: &str) -> Result<&'a Node, LookupError> {
```
InsertCreativityHere, line 223: Helper function to pull some common logic out of our `find` functions.

slicec/src/patchers/comment_link_patcher.rs
```diff
@@ -101,13 +102,16 @@ impl CommentLinkPatcher<'_> {
...
+                if Primitive::from_str(&identifier.value).is_ok() {
+                    "primitive types cannot be linked to".to_owned()
+                } else {
+                    assert!(matches!(lookup_error, LookupError::DoesNotExist { .. }));
+                    format!("no element named '{}' exists in scope", identifier.value)
+                }
             })
```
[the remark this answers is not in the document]
InsertCreativityHere: This is actually correct behavior. Writing `::int32` is **NOT** a reference to the primitive type `int32`. In that case you are specifically asking to link against the user-defined module like ` module \int32`. If it doesn't exist, then "no element named..." is correct. If it does exist, you will get "modules cannot be linked to". Either way, this is correct.

slicec/src/ast/mod.rs
```diff
@@ -271,6 +219,15 @@ impl Ast {
...
```
pepone: Not sure if this is a rust convention, but why not use option, seems the only error is `DoesNotExists` which can be represent by returning `None`.

pepone (commented):
I think this can be simplified:

* Use a single lookup mechanism. Splitting `find_primitive_node` from `find_node_by_id` gives primitives special lookup semantics. Primitives should participate in the same scoped lookup as other types.

* Don’t use `\` to select between a primitive and a user-defined type. Treat primitives as predefined types in the global type namespace and apply normal shadowing. Under the current lexer, a declaration may still need to be written as `struct \string {}`, but an unqualified `string` reference in that scope should resolve to the struct, while `::string` should resolve to the global primitive.

* Return `Option` from the low-level `find`. Name lookup itself has only one failure condition: nothing was found. A typed conversion may still return `Result` when distinguishing a type mismatch is useful, but `DoesNotExist` is better represented by `None`.

This would also give type references and documentation links the same resolution semantics and avoid the enum-index-based primitive lookup introduced by this PR.

InsertCreativityHere:
Closing this until we fully settle on changing the mechanics of the primitive types and type-resolution as a whole.
All of this is now targeted for `0.5.0`; none of these bugs are critical to fix in `0.4.1` anyways.

Issue #625, opened 2023-07-03
Update Comment Validators to Check for Duplicate Tags


InsertCreativityHere:
For example, things like these are let through currently, when they should not be:
```
@param foo: my comment
@param foo: duplicate comment
```
or something less obvious but still incorrect:
```
@returns: Something
@returns foo: Something else
```

PR #818, opened 2026-09-01
Fix Forwarding of Doc-Comments on Return-Parameters and Variant-Fields

### LINKED ISSUE #809: `@returns` and variant-field `@param` doc comments never reach code generators
In the `Compiler` module, operation parameters, operation return fields, and variant enum fields are all represented as `Field`, which holds an `EntityInfo`, which in turn holds a `comment: DocComment?` field.

In Slice, doc comments cannot be written on these fields directly — they are written as `@param`/`@returns` tags on the enclosing operation or variant. `slice_file_converter.rs` bridges this for **operation parameters**: `get_doc_comment_for_parameter` finds the matching `@param` tag on the enclosing operation and synthesizes the parameter's `entityInfo.comment` from the tag's message, so code generators see parameter documentation without dealing with tag mechanics.

The other two `Field` contexts never get their `entityInfo.comment` filled:

- **Return fields**: `convert_operation` converts the fields of `returnType` with the same `convert_parameter` used for parameters, but `get_doc_comment_for_parameter` searches only `comment.params` — `comment.returns` is never consulted, so the comment stays unset and `@returns` messages are parsed, validated, and then dropped. Note `ReturnsTag.identifier` is optional: an unnamed `@returns` documents the operation's single return field.

- **Variant enum fields**: the comment validator accepts `@param` on enumerators (`only_operations_have_parameters` allows `Entities::Enumerator`), but `convert_field` reads only the field's own comment, which inline variant fields never have — the variant's `@param` tags are dropped, so the positional record parameters generated for these fields have no documentation.

Both need the same treatment as operation parameters: synthesize the field's `entityInfo.comment` from the enclosing entity's tags. No change to the `Compiler` Slice definitions is required.

InsertCreativityHere:
This PR fixes #809, by ensuring that we detect and forward `@returns` for return parameters, and `@param` for fields of enumerators. Previously, these were not detected at all, and were accidentally being thrown away.

The fix for parameters was to add a `is_return: bool` that we check to know whether we're searching for `@param` or `@returns`.
We could split this into 2 functions instead of using a trailing bool; the function is small, and shares some code. But if anyone prefers 2 functions instead of my little bool, feel free to say so.

The fix for enumerator fields, is we re-use the logic for creating a struct field, but then 'patch' it's doc-comment with any `@param` that were present on the enumerator.

## What's Changed

**slicec**:
- The fields of a `Variant` now correctly have their doc-comments populated by `@param` tags on their enclosing variant.
- The `returnType` fields of an `Operation` now correctly have their doc-comments populated by `@returns` tags on their enclosing operation

### DIFF
--- a/slicec/src/slice_file_converter.rs
+++ b/slicec/src/slice_file_converter.rs
@@ -269,7 +269,7 @@ fn get_entity_info_for(element: &impl Commentable) -> EntityInfo {
     }
 }
 
-/// Returns a [DocComment] with it's overview set to the provided message, with no other tags.
+/// Returns a [DocComment] with its overview set to the provided message, with no other tags.
 fn get_basic_doc_comment_from_message(message: &slicec::grammar::Message) -> DocComment {
     DocComment {
         overview: message.value.iter().map(Into::into).collect(),
--- a/slicec/src/slice_file_converter.rs
+++ b/slicec/src/slice_file_converter.rs
@@ -273,20 +273,31 @@ fn get_entity_info_for(element: &impl Commentable) -> EntityInfo {
 ///
 /// In Slice, doc-comments are not allowed on parameters. Instead, you would use a '@param' tag applied to an enclosing
 /// operation. But this is an implementation detail of the language, not something code-generators should deal with.
-fn get_doc_comment_for_parameter(parameter: &GrammarParameter) -> Option<DocComment> {
+fn get_doc_comment_for_parameter(parameter: &GrammarParameter, is_return: bool) -> Option<DocComment> {
     let operation_comment = parameter.parent().comment()?;
 
-    // We get the parameter's doc-comment in 3 steps:
-    // 1) Try to find a matching '@param' tag on the operation's doc-comment.
-    // 2) If one was present, extract just its `Message` field, and convert it to the mapped type.
-    // 3) Construct a mapped `DocComment` which contains the mapped message.
-    operation_comment.params.iter()
-        .find(|param_tag| param_tag.identifier.value == parameter.identifier())
-        .map(|param_tag| param_tag.message.value.iter().map(Into::into).collect())
-        .map(|message| DocComment {
-            overview: message,
-            see_tags: Vec::new(),
-        })
+    let message = if is_return {
+        let return_tag = if parameter.identifier() == "returnValue" {
+            // If this is a single return-type, we search the operation's doc-comment for a '@returns' tag with no name.
+            operation_comment.returns.iter().find(|return_tag| return_tag.identifier.is_none())
+        } else {
+            // If this is a named return parameter, we search the operation's doc-comment for a matching '@returns' tag.
+            operation_comment.returns.iter().find(|return_tag| {
+                return_tag.identifier.as_ref().map(|id| id.value.as_str()) == Some(parameter.identifier())
+            })
+        };
+        return_tag.map(|return_tag| &return_tag.message)
+    } else {
+        // If this isn't a return parameter, we search the operation's doc-comment for a matching '@param' tag.
+        operation_comment.params.iter()
+            .find(|param_tag| param_tag.identifier.value == parameter.identifier())
+            .map(|param_tag| &param_tag.message)
+    };
+
+    message.map(|m| DocComment {
+        overview: m.value.iter().map(Into::into).collect(),
+        see_tags: Vec::new(),
+    })
 }
 
 /// Helper function to convert the result of `tag.linked_entity()` into an [`EntityId`].
@@ -426,7 +437,7 @@ impl SliceFileContentsConverter {
     fn convert(contents: &[GrammarDefinition]) -> Vec<Symbol> {
         // Create a new converter.
         let mut converter = SliceFileContentsConverter {
-            converted_contents: Vec::new()
+            converted_contents: Vec::new(),
         };
 
         // Iterate through the provided file's contents, and convert each of it's top-level definitions.
@@ -457,11 +468,11 @@ impl SliceFileContentsConverter {
         Struct {
             entity_info: get_entity_info_for(struct_def),
             is_compact: struct_def.is_compact,
-            fields: struct_def.fields().into_iter().map(|e| self.convert_field(e)).collect(),
+            fields: struct_def.fields().into_iter().map(|e| self.convert_struct_field(e)).collect(),
         }
     }
 
-    fn convert_field(&mut self, field: &GrammarField) -> Field {
+    fn convert_struct_field(&mut self, field: &GrammarField) -> Field {
         Field {
             entity_info: get_entity_info_for(field),
             tag: field.tag.as_ref().map(|integer| integer.value as i32),
@@ -483,12 +494,12 @@ impl SliceFileContentsConverter {
         Operation {
             entity_info: get_entity_info_for(operation),
             is_idempotent: operation.is_idempotent,
-            parameters: operation.parameters().into_iter().map(|e| self.convert_parameter(e)).collect(),
+            parameters: operation.parameters().into_iter().map(|e| self.convert_parameter(e, false)).collect(),
             has_streamed_parameter: operation
                 .parameters
                 .last()
                 .is_some_and(|parameter| parameter.borrow().is_streamed),
-            return_type: operation.return_members().into_iter().map(|e| self.convert_parameter(e)).collect(),
+            return_type: operation.return_members().into_iter().map(|e| self.convert_parameter(e, true)).collect(),
             has_streamed_return: operation
                 .return_type
                 .last()
@@ -496,11 +507,11 @@ impl SliceFileContentsConverter {
         }
     }
 
-    fn convert_parameter(&mut self, parameter: &GrammarParameter) -> Field {
+    fn convert_parameter(&mut self, parameter: &GrammarParameter, is_return: bool) -> Field {
         let parameter_info = EntityInfo {
             identifier: parameter.identifier().to_owned(),
             attributes: get_attributes_from(parameter.attributes()),
-            comment: get_doc_comment_for_parameter(parameter),
+            comment: get_doc_comment_for_parameter(parameter, is_return),
         };
 
         Field {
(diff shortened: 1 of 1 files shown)


slicec/src/slice_file_converter.rs
```diff
@@ -540,14 +551,31 @@ impl SliceFileContentsConverter {
...
+        let mut converted_field = self.convert_struct_field(field);
+
+        // Variant fields get '@param' tags on their parent enumerator, so if there is a comment on the enumerator,
+        // we check for a matching '@param' tag, and use its message as the field's comment.
+        if let Some(variant_comment) = enumerator.comment() {
+            converted_field.entity_info.comment = variant_comment.params.iter()
+                .find(|param_tag| param_tag.identifier.value == field.identifier())
+                .map(|param_tag| param_tag.message.value.iter().map(Into::into).collect())
+                .map(|overview| DocComment { overview, see_tags: Vec::new() });
+        }
```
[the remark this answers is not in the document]
InsertCreativityHere: We aren't concerned about users placing doc-comments directly on enumerator fields since we're going to remove this functionality. If there are conflicting comments on both the enumerator and it's field, it's fine that the enumerator one 'wins'.

pepone (approved):
LGTM.

**Suggestion (non-blocking)**
- A regression test here would be the first for `slice_file_converter.rs`, but needs no new infrastructure: a `#[cfg(test)] mod` in the binary crate that compiles a snippet with `compile_from_strings`, converts it with `SliceFile::from`, and asserts on `Field.entity_info.comment`. The module below fails 3/4 on `main` and passes on this branch.

<details>
<summary><code>slicec/src/converter_doc_comment_tests.rs</code> (wire in with <code>#[cfg(test)] mod converter_doc_comment_tests;</code> in <code>main.rs</code>)</summary>

```rust
//! Tests that `@param` / `@returns` doc-comment tags are forwarded onto the mapped `Field`s that code-generator
//! plugins receive (operation parameters, return members, and variant fields).
//
// To wire in: add `#[cfg(test)] mod converter_doc_comment_tests;` next to `mod slice_file_converter;` in main.rs.

use crate::definition_types::*;
use slicec::compile_from_strings;

/// Compiles `slice` (asserting no errors) and returns the converted contents of the file.
fn convert(slice: &str) -> Vec<Symbol> {
    let state = compile_from_strings(&[slice], None);
    assert!(!state.diagnostics.has_errors(), "{:?}", state.diagnostics);
    SliceFile::from(&state.files[0]).contents
}

/// Flattens a mapped doc-comment's overview into a single string (links rendered as `{@link id}`).
fn overview_of(field: &Field) -> Option<String> {
    field.entity_info.comment.as_ref().map(|comment| {
        comment.overview.iter().map(|component| match component {
            MessageComponent::Text(text) => text.clone(),
            MessageComponent::Link(link) => format!("{{@link {link}}}"),
        }).collect()
    })
}

fn operations(symbols: &[Symbol]) -> &[Operation] {
    let Some(Symbol::Interface(interface)) = symbols.iter().find(|s| matches!(s, Symbol::Interface(_))) else {
        panic!("no interface in {symbols:?}");
    };
    &interface.operations
}

fn variants(symbols: &[Symbol]) -> &[Variant] {
    let Some(Symbol::VariantEnum(variant_enum)) = symbols.iter().find(|s| matches!(s, Symbol::VariantEnum(_))) else {
        panic!("no variant enum in {symbols:?}");
    };
    &variant_enum.variants
}

#[test]
fn param_tags_are_forwarded_to_parameters() {
    let symbols = convert("
        module tests
        interface I {
            /// @param p: the param
            op(p: string, undocumented: int32)
        }
    ");
    let op = &operations(&symbols)[0];
    assert_eq!(overview_of(&op.parameters[0]), Some("the param\n".to_owned()));
    assert_eq!(overview_of(&op.parameters[1]), None);
}

#[test]
fn unnamed_returns_tag_is_forwarded_to_single_return_type() {
    let symbols = convert("
        module tests
        interface I {
            /// @returns: the value
            plain() -> int32

            /// @returns: the stream
            streamed() -> stream int32

            /// @returns: the tagged value
            tagged() -> tag(1) int32?
        }
    ");
    for op in operations(&symbols) {
        assert_eq!(op.return_type.len(), 1, "{}", op.entity_info.identifier);
        assert!(overview_of(&op.return_type[0]).is_some(), "{}", op.entity_info.identifier);
    }
}

#[test]
fn named_returns_tags_are_forwarded_to_return_tuple_members() {
    let symbols = convert("
        module tests
        interface I {
            /// @returns b: second
            /// @returns a: first
            op() -> (a: int32, b: string, c: bool)
        }
    ");
    let op = &operations(&symbols)[0];
    assert_eq!(overview_of(&op.return_type[0]), Some("first\n".to_owned()));
    assert_eq!(overview_of(&op.return_type[1]), Some("second\n".to_owned()));
    assert_eq!(overview_of(&op.return_type[2]), None);
}

#[test]
fn param_tags_are_forwarded_to_variant_fields() {
    let symbols = convert("
        module tests
        enum E {
            /// Variant overview.
            /// @param y: the y field
            A(x: int32, y: string)
            B(z: int32)
        }
    ");
    let variants = variants(&symbols);
    assert_eq!(overview_of(&variants[0].fields[0]), None);
    assert_eq!(overview_of(&variants[0].fields[1]), Some("the y field\n".to_owned()));
    assert_eq!(overview_of(&variants[1].fields[0]), None);
}
```

</details>

**Author's call**
- `parameter.parent().return_members().len()` collects a `Vec` just to count it, once per return member; `parameter.parent().return_type.len()` is the same check without the allocation, and matches how `convert_operation` already reads `return_type` directly.
- The `is_return` bool is fine as-is.

**Follow-ups, not for this PR:** an unnamed `@returns:` on a tuple-returning operation passes validation without a lint and, with no single member to attach it to, never reaches a generator either (#819). Same shape for an enumerator `@param` whose name matches no field, or that sits on a basic enum (#820).

bernardnormier (commented):
LGTM, with a few small items not already raised by pepone or Copilot.

**Nit**
- `slice_file_converter.rs:274`: "you would use" → "you use".

**Reuse (author's call)**
- The grammar `Message` → mapped overview conversion is now written three times in this file (`get_doc_comment_for_parameter` at ~L298, the `From<&GrammarDocComment>` impl at ~L392, and `convert_variant_field` at ~L569), and the `@param` find-and-map chain twice (~L291-300 and ~L566-570). An `impl From<&GrammarMessage> for DocComment` next to the existing `From<&GrammarDocComment>` impl, plus a small `find_param_message(params, identifier)` helper, would collapse all of them, and would let `convert_variant_field` set the comment when constructing the field instead of building it via `convert_struct_field` and patching it afterwards.

**Follow-up, not for this PR**
- No validator rejects duplicate `@returns` (or `@param`) tags for the same member, and both lookups use `find`, so only the first message is forwarded and the rest are silently dropped. Pre-existing for `@param`, newly applicable to `@returns` now that they are consumed.

InsertCreativityHere:
@pepone

> A regression test here would be the first for slice_file_converter.rs

I did add a test originally, but it ended up not being very small or very unit-testy, so I removed it.
Right now we don't have great infrastructure for testing the binary (as opposed to the library); we should add this
in the future. Also, much of this hand-written code is going to be replaced by Rust generated code in 0.5.0, which
will hopefully not be too long after releasing 0.4.1. It's for all these reasons I ended up not committing a test.

> parameter.parent().return_members().len() collects a Vec just to count it

My Claude also flagged this minor inefficiency, but I think what the PR does is correct.
Right now, `slicec` has many fields that are public, but probably should not be, and `return_type` is one of these.
It's more correct to access this information through the `return_members()` function, instead of directly as a field.
One day, I will go through and make many of these correctly private.

> Follow-ups, not for this PR...

The first point about unnnamed `@returns` is I believe, intentional design, in the case you want to document the whole return type, instead of each individual field. Might very well still be broken though.

The second point is definitely a bug. Will fix separately.

InsertCreativityHere:
@bernardnormier

I updated the wording to "you use"

I introduced a helper function for the conversion of `Message` -> `overview only DocComment`.
But it's a standalone function, not a `From`. Since `From` kind of implies this is the de-facto conversion.
And really, converting a single overview to a full `DocComment` doesn't feel right for this.

And, for your duplication findings, this is a known issue, that I'd like to fix for 0.4.1; see: #625

bernardnormier (commented):
Thanks for the fixes. One more thing before approving: several of the comments in this PR narrate the code right below them, which is exactly what we want to avoid.

**Please remove**
- L566 `// Construct the field as normal by re-using the logic of 'convert_struct_field'.` — the next line is the call to `convert_struct_field`.
- L300 `// If this isn't a return parameter, we search the operation's doc-comment for a matching '@param' tag.` — this is the `else` branch, spelled out.
- L444 `// Iterate through the provided file's contents, and convert each of its top-level definitions.` — pre-existing, but since you touched it: it describes a `for` loop over `contents`.

**Please trim to the "why"**
- L289-290: the two sentences mirror the two arms of the `match`. The only non-obvious fact is the language rule: a single return type is documented by an unnamed `@returns`, tuple members by name. One sentence stating that rule is enough.
- L569-570: "variant fields get '@param' tags on their parent enumerator" is the why and worth keeping. The rest ("we check for a matching tag and use its message as the field's comment") restates the `find`/`map`.

The doc-comments on the new helper and on `get_doc_comment_for_parameter` are fine: they describe the API and the language rule, not the implementation. And deleting the old "3 steps" comment was the right call, same problem.

slicec/src/slice_file_converter.rs
```diff
@@ -378,15 +394,15 @@ impl From<&GrammarSliceFile> for SliceFile {
 impl From<&GrammarDocComment> for DocComment {
     fn from(doc_comment: &GrammarDocComment) -> Self {
         let overview = doc_comment.overview.as_ref().map(|message| {
-            message.value.iter().map(Into::into)
+            message.value.iter().map(Into::into).collect()
```
bernardnormier, line 397: This is the same `Message` -> overview conversion that `get_basic_doc_comment_from_message` does. The helper can't be used here as-is because it returns a whole `DocComment` and this impl needs the bare overview plus real `see_tags`.

Suggest extracting the message conversion itself, e.g. `fn convert_message(message: &slicec::grammar::Message) -> Message { message.value.iter().map(Into::into).collect() }`, and calling it from both `get_basic_doc_comment_from_message` and here:

```rust
let overview = doc_comment.overview.as_ref().map(convert_message);
```

bernardnormier (commented):
One more on the helper, to finish the de-duplication.

slicec/src/slice_file_converter.rs
InsertCreativityHere: I'm not convinced this extra helper earns it's keep. It's a single-line of code, only called from 2 places.
I added it, and it actually increases our net-line-count by 3 lines.

I'd be fine with dropping your other helper (`get_basic_doc_comment_from_message`) in place of the `convert_message` suggestion. But if we added both helpers, then we'd have 2 helpers which are collectively only called by 3 places (that's together, not each). Only 1.5 calls per helper function is below the bar of usefulness IMO.
If the implementations were longer, I could see the point, but both helpers are basically one-liners.

InsertCreativityHere:
Of the 5 doc-comments you flagged:
- I deleted the 2 comments which were obvious narration with no other value.
- I reworked the comment about "variant fields and `@param`". I agree about the comment including the fact we search for `@param` on the parent element, but I also think it's useful to state that it isn't just about copying this tag, but that we 'promote' the tag's description to a full-blown doc-comment itself. This is what the comment was trying to express.

- And for the remaining comments in `get_doc_comment_for_parameter`:

> The only non-obvious fact is the language rule: a single return type is documented by an unnamed @returns, tuple members by name

I feel the opposite here. The fact that single return types lack a name is, IMO, an obvious fact of the Slice language.
What I find non-obvious is that when we transmit these parameters over the wire, we synthesize a doc-comment for them
based on the `@param` and `@returns` tags on their parent operation. This is non-obvious because that's not how they're
expressed by the Slice language, nor in `slicec` itself.

bernardnormier (approved):

Issue #684, opened 2023-12-11
Use @param to add doc-comment to enumerator fields


bernardnormier:
We should use @param to add doc-comments to the fields of an enumerator.

InsertCreativityHere:
On a high-scale level it feels like we're trying to make enums with fields into enums with parameters.
We use the parameter parenthesis `(...)` instead of field braces `{...}`, the `@param` tag instead of direct comment placement.

Should we just embrace this and rename our construct to "enum with parameters"?

In Rust there is truly no difference between fields, whether they're in a struct or in a variant (ie. enumerator),
which I find to be very nice and clean, compared to ours.

bernardnormier:
We should keep the name "enum with fields". They are fields. This is very similar to records in C#, where you can specify your properties with a primary constructor:

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record

> When you declare a [primary constructor](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/instance-constructors#primary-constructors) on a record, the compiler generates public _properties_ for the primary constructor parameters.

InsertCreativityHere:
This was partially implemented by https://github.com/icerpc/slicec/commit/4838a71b6235621e3a7ad8f07bb6a13a13986fff

But we still need to add validation and tests for it.

InsertCreativityHere:
This is finally complete, as of #822.

PR #822, opened 2026-09-10
Validate that '@param' comments on enumerators match a real field of that enumerator.

### LINKED ISSUE #820: Enumerator `@param` tags are never validated against the enumerator's fields
`@param` tags are permitted on enumerators (`only_operations_have_parameters` in `slicec/src/validators/comments.rs` allows `Entities::Enumerator`), but unlike operation parameters (`validate_param_tags` in `slicec/src/validators/operations.rs`) the tag's identifier is never checked against the enumerator's fields, and nothing rejects the tag on an enumerator that has no fields. All three of these compile without a single lint:

```slice
module tests

enum E {
    /// @param nmae: typo, no field is called 'nmae'
    A(name: string)

    /// @param x: this enumerator has no fields
    B
}

enum Status : int32 {
    /// @param code: enumerators of a basic enum can never have fields
    Ok = 0
}
```

With #818, a variant field only receives a comment when a `@param` tag names it, so the typo above silently produces an undocumented field in the generated code; the other two tags are silently discarded.

Suggested fix: validate enumerator `@param` identifiers against `enumerator.fields()` the same way `validate_param_tags` does for operations. That also rejects `@param` on fieldless and basic-enum enumerators.

---
Found while reviewing #818 (Claude Fable 5.1, cross-checked with Codex; the silent acceptance was confirmed by compiling the snippet above and observing zero diagnostics). Not yet human-triaged.

InsertCreativityHere:
Currently `slicec` doesn't validate that `@param` tags on enumerators are for real fields. For example, this produces no lints:
```
enum Foo {
    /// @param fakename: this field doesn't exist.
    A(hello: int32)
}
```

This PR directly fixes #820, and is also the last piece of validation we were missing to consider #684 fully implemented.

----

What's Changed:
- slicec: `slicec` now validates that `@param` tags on enumerators reference an existent field of the enumerator, like it already did for operation parameters.


slicec/tests/comment_tests.rs
```diff
@@ -513,7 +574,7 @@ mod comments {
     }
 
     #[test]
-    fn param_tags_can_only_be_used_with_operations() {
+    fn param_tags_are_rejected_on_incorrect_elements() {
```
pepone, line 577: Could we update this expected message and the corresponding diagnostic in validators/comments.rs? Enumerators are now valid owners of @param tags, so 'only operations can have parameters' is misleading. For example: 'comment has a param tag, but only operations and enumerators can use param tags'.

pepone (approved):

pepone (approved):

externl (approved):

slicec/tests/comment_tests.rs
InsertCreativityHere: Updated the message, and the names of the validation functions around it!

PR #824, opened 2026-09-11
Reject Streamed Parameters that are Also Tagged

### LINKED ISSUE #823: `slicec` Does not Reject Tagged Streamed Parameters
Tags are all about adding and removing parameters/fields while maintaining on-the-wire compatibility.
Steamed parameters can always be added/removed without affecting on-the-wire compatibility.

Therefore tagging a streamed parameter never makes sense to do. It's not necessary, and streams use their own encoding, distinct from tags anyways.

`slicec` should of been rejecting these from the get-go, and the fact it allows `tag(6) g: stream int32?` at all is an oversight.

## What's Changed

`slicec` now rejects tags on streamed parameters with an error.

InsertCreativityHere:
This PR fixes #823, which I realized while reviewing the recent docs changes to streamed parameters.

Apparently, `slicec` has never rejected these: `tag(5) myParam: stream int32?` - which it should.
Tags allow you to add/remove parameters and fields while maintaining on-the-wire compatibility, which is already possible for `stream`ed parameters, even without `tag`s. So this was never a meaningful thing to write.

Additionally, `stream`s use their own encoding and semantics, both different from `tag`s.

----

I'm targeting this for 0.4.1, even though it's technically a source-incompatible change. I don't think anyone is doing this, it's never been an _intentionally_ supported syntax, and it's a trivial change (just remove the `tag(6)`) for users.

For the record, the code generators just completely ignore the tag in this case, since it splits parameters into streamed and non-streamed before checking `tag`s.

## What's Changed:
slicec
- `slicec` now emits an error for streamed parameters that also have tags. There is no point in applying a tag to a streamed parameter, and any such tags should be removed.


slicec/tests/interfaces/operations.rs
```diff
@@ -315,4 +315,25 @@ mod streams {
...
     }
+
+    #[test]
+    fn stream_parameter_cannot_be_tagged() {
+        // Arrange
+        let slice = "
+            module Test
+
+            interface I {
+                op(tag(79) s: stream bool?)
```
pepone: Maybe add a second test case for stream return?

pepone (approved):

slicec/src/diagnostics/errors.rs
```diff
@@ -99,6 +99,12 @@ pub enum Error {
         parameter_identifier: String,
     },
 
+    /// A streamed parameter was also tagged.
```
externl, line 102:
```suggestion
    /// A streamed parameter must not be tagged.
```

externl (approved):

slicec/tests/interfaces/operations.rs
InsertCreativityHere: Added a 2nd test case. We have one operation which tests an input-parameter, and a second operation which tests a return-type.

slicec/src/diagnostics/errors.rs
InsertCreativityHere: I think the current wording is consistent with the rest of the file. We structure our errors like:
```
/// What actually happened
TheRuleThatWasBroken,
```
Otherwise, it'd just be like restating the same thing twice.

Here, the variant is named `StreamedParamsCannotBeTagged` (the rule),
and the comment says"A streamed parameter was also tagged" (what concretely happens to break the rule).

PR #825, opened 2026-09-14
Sign the Windows slicec binary in release builds

pepone:
The release workflow uploads `slicec.exe` to download.zeroc.com without an Authenticode signature, and `IceRpc.Slice.Tools` bundles it unchanged. The `tools/windows-x64/slicec.exe` in `IceRpc.Slice.Tools` 0.6.0 on nuget.org is `NotSigned`, while the managed assemblies in the same package carry ZeroC's signature. This was found while fixing icerpc/icerpc-csharp#4949.

This PR signs the Windows binary on the Windows build leg, before the artifact upload, with the same Trusted Signing account, certificate profile and action version as icerpc-csharp's `build-packages.yml`:

- **Releases (any quality other than `nightly`):** signing is mandatory.
- **Nightly builds:** signing is optional, through a new `authenticode_sign` input (default `false`) on `workflow_call` and `workflow_dispatch`. The scheduled nightly doesn't set it, so it stays unsigned. A manual nightly run can opt in.
- **Verify step:** after signing, a step fails the build unless `Get-AuthenticodeSignature` reports `Valid`, so a signing problem can't lead to an unsigned binary being uploaded. The `upload` job needs `build`, so nothing is published in that case.

The signing action is pinned to a commit SHA, as in icerpc-csharp, because it receives the signing credentials.

## Before merging

The `AZURE_TENANT_ID`, `AZURE_CLIENT_ID` and `AZURE_CLIENT_SECRET` secrets are organization secrets that are not currently shared with this repository. An org admin needs to add `icerpc/slicec` to their selected repositories. Until then, a non-nightly release fails at the signing step. The scheduled nightly doesn't sign, so it isn't affected.

## Testing

- `actionlint` 1.7.12 reports no issues.
- The signing policy script gives `true` for `stable` and `RC` with any input value, `false` for `nightly` by default, and `true` for `nightly` with `authenticode_sign: true`.
- The verify step's PowerShell passes on a ZeroC-signed assembly from nuget.org and fails on the unsigned 0.4.0 `slicec.exe`.
- The signing step itself can only run here once the secrets are shared. A manual `workflow_dispatch` with `quality: nightly` and `authenticode_sign: true` exercises it without publishing a stable release.

The workflow is the same on `0.4.x`, so this commit cherry-picks cleanly and a 0.4.1 release would ship a signed `slicec.exe`.

## What's Changed

None - release pipeline change: the published Windows `slicec` binary is now Authenticode-signed; the compiler itself is unchanged.


externl (approved):

.github/workflows/build-release.yml
```diff
@@ -44,6 +54,20 @@ jobs:
           echo "version=$version" >> "$GITHUB_OUTPUT"
           echo "channel=$channel" >> "$GITHUB_OUTPUT"
 
+      - name: Get Signing Policy
```
InsertCreativityHere: Not a CI expert here, so maybe there's a reason we can't do this, but...

Do we need a whole step for this? All it's doing is setting an environment variable which is read later on in `Sign slicec binary`.

Couldn't we delete this `Get Signing Policy` step, and then just update the `if` in `Sign slicec binary` to check `inputs.authenticode_sign || inputs.quality != 'nightly'`?

----

At the very least right now the check feels split. In this part we do `if !nightly || authenticode_sign`, and in the later parts where we actually use it, we also check for `windows`. I just feel like this check should all be done in one place, instead of being split half and half.

.github/workflows/build-release.yml
```diff
@@ -78,6 +102,32 @@ jobs:
...
+      # Fail the build rather than upload an unsigned binary when signing is required.
+      - name: Verify slicec binary signature
+        if: runner.os == 'Windows' && needs.metadata.outputs.authenticode_sign == 'true'
+        shell: pwsh
+        run: |
+          $signature = Get-AuthenticodeSignature "target\${{ matrix.target }}\release\${{ matrix.artifact }}"
+          if ($signature.Status -ne 'Valid') {
+            throw "${{ matrix.artifact }} is not validly signed: $($signature.Status) $($signature.StatusMessage)"
+          }
+          Write-Output "Signed by $($signature.SignerCertificate.Subject)"
```
InsertCreativityHere, line 123: My Claude had this to say, I'm not familiar enough with Authenticode signing unfortunately:

----

Require a timestamp in the verify step. Artifact Signing certificates are renewed daily and are valid for only 72 hours. Without an RFC 3161 countersignature, a signature becomes invalid three days later, so an untimestamped binary would pass the current Status -eq 'Valid' check on the runner and then break for users. I could not confirm what the v2 client does if the timestamp server is down, so this check is cheap insurance. Tested locally against a signed and an unsigned binary:
```
$signature = Get-AuthenticodeSignature "target\${{ matrix.target }}\release\${{ matrix.artifact }}"
if ($signature.Status -ne 'Valid') {
  throw "${{ matrix.artifact }} is not validly signed: $($signature.Status) $($signature.StatusMessage)"
}
if (-not $signature.TimeStamperCertificate) {
  throw "${{ matrix.artifact }} is signed but not timestamped; the signature would expire with the 72-hour certificate."
}
Write-Output "Signed by $($signature.SignerCertificate.Subject)"
```

InsertCreativityHere (approved):
Looks good, maybe some things could be simplified, maybe not.

One tangential ask, as someone who does test builds off their fork of `slicec`, could we also check `github.repository == 'icerpc/slicec'` so the new actions only run for our repository?

Without that, forks will always have red CI, since they can't access our secrets, and so the job would fail.

.github/workflows/build-release.yml
pepone: Applied in 213c3ca. Removed the Get Signing Policy step and its metadata output. Both signing and verification now use the full condition directly: github.repository == 'icerpc/slicec' && runner.os == 'Windows' && (inputs.authenticode_sign || inputs.quality != 'nightly'). This also addresses your request to skip signing on forks.

.github/workflows/build-release.yml
pepone: Applied in 213c3ca. Verification now requires TimeStamperCertificate in addition to Status == 'Valid', so a currently valid but untimestamped binary fails before artifact upload. The RFC 3161 timestamp settings remain in the signing step. Validated the extracted PowerShell with mocked valid/timestamped, valid/untimestamped, unsigned, and invalid/timestamped results; all four cases behaved as expected. Actual Azure signing still needs a Windows workflow run.

pepone:
Addressed all three review suggestions in 213c3ca:

- Removed the signing-policy step and output; signing and verification now check the full condition directly.
- Added `github.repository == 'icerpc/slicec'` to both steps, so fork builds skip signing and verification even for non-nightly releases or an explicit signing opt-in.
- Required a timestamp certificate as well as a valid signature before artifact upload.

Validation: actionlint 1.7.12 and git diff --check pass; all 36 combinations of repository, OS, release quality, and signing opt-in pass; the extracted PowerShell verification script parses and passes four mocked signature cases. Actual Azure signing was not exercised locally.

Replied to and resolved both inline review threads.

Issue #800, opened 2026-08-19
Doc-Comment Indent Stripping Splits a UTF-8 Codepoint


InsertCreativityHere:
`comments/grammar.rs:105–125`: the common leading-whitespace amount is computed as a byte offset via `str::find`, min'd across lines, then applied to every line with `replace_range(..common, "")`.

If lines have different-byte-width leading whitespace (an ASCII space on one line, a multi-byte Unicode space like NBSP U+00A0 on another — a common copy-paste artifact), the byte offset lands mid-codepoint → panic `end of range should be a character boundary`. Parsed at construction time, before validation:

## Reproduction

```
printf '/// a\n///\xc2\xa0b\nstruct S {}\n' > nbsp.slice && ./slicec nbsp.slice --dry-run
```

It's also a latent correctness bug (mixed indentation stripped by the wrong amount).

## Fix

Measure and strip in chars, not bytes: `char_indices().nth(common)`.

PR #828, opened 2026-09-14
Pre-Flatten Comment Messages in the Parser Rules

InsertCreativityHere:
This PR makes a slight simplification to the grammar rules used to parse doc-comment messages, which I noticed while working on #800.

## What's Changed

None - This PR is just an internal simplification of the grammar rules used to parse doc-comments with no side-effects.


slicec/src/parsers/comments/grammar.lalrpop
```diff
@@ -80,20 +80,20 @@ InlineLink: LinkTag = {
...
 
 MessageLines: Message = {
-    <l: @L> <m: (<Message?> newline)+> <r: @R> => {
+    <l: @L> <m: (<Message> newline)+> <r: @R> => {
         let span = Span::new(l, r, comment_parser.file_name);
         sanitize_message_lines(m, span)
     }
 }
 
-Message = MessageComponent+;
```
InsertCreativityHere, line 96: At a grammar level, these two constructs should be equivalent:
`(MessageComponent+)?` (match 1 more components, but also accept 0 components)
`MessageComponent*` (match 0 or more components)

But in addition to being a simpler rule, it also simplifies the types.
The 1st expression gives `Option<Vec<...>>`, whereas the 2nd gives just `Vec<...>`.

